PR #1, opened 2018-11-30
Initial empty C# micro-generator project

chrisdunelm:


chrisdunelm:
Those files are already there. I admit that I pushed them directly to master!

PR #4, opened 2018-12-05
Initial code generation

chrisdunelm:
Just enough implemented to generate the structure of the client Settings class, with a test.
Code formatting not yet implemented, which is why the test can't use the Roslyn parser yet to check the result.

### DIFF
--- a/Google.Api.Generator/Generation/SourceFileContext.cs
+++ b/Google.Api.Generator/Generation/SourceFileContext.cs
@@ -1,0 +1,148 @@
+﻿// Copyright 2018 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using Google.Api.Generator.Utils;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using static Microsoft.CodeAnalysis.CSharp.SyntaxFactory;
+
+namespace Google.Api.Generator.Generation
+{
+    internal class SourceFileContext
+    {
+        public enum ImportStyle
+        {
+            FullyAliased,
+            // TODO: Unaliased
+        }
+
+        private static readonly IReadOnlyDictionary<string, string> s_wellknownNamespaceAliases = new Dictionary<string, string>
+        {
+            { typeof(System.Int32).Namespace, "sys" }, // Don't use "s"; one-letter alias cause a compilation error!
+            { typeof(System.Collections.Generic.IEnumerable<>).Namespace, "scg" },
+            { typeof(Google.Api.Gax.Expiration).Namespace, "gax" },
+            { typeof(Google.Api.Gax.Grpc.CallSettings).Namespace, "gaxgrpc" },
+            { typeof(Grpc.Core.CallCredentials).Namespace, "grpccore" },
+            { typeof(Grpc.Core.Interceptors.Interceptor).Namespace, "grpcinter" },
+            { typeof(Google.Protobuf.WellKnownTypes.Any).Namespace, "wkt" },
+        };
+
+        public SourceFileContext(ImportStyle importStyle) => _importStyle = importStyle;
+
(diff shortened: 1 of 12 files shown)


Google.Api.Generator/Generation/SourceFileContext.cs
```diff
@@ -0,0 +1,148 @@
...
+                name = $"{declaringTyp.Name}.{name}";
+                declaringTyp = declaringTyp.DeclaringTyp;
+            }
+            SimpleNameSyntax result = IdentifierName(name);
+            if (typ.GenericArgTyps != null)
+            {
+                // Generic typ, so return a generic name by recursively calling this method on all type args.
+                result = GenericName(result.Identifier, TypeArgumentList(SeparatedList(typ.GenericArgTyps.Select(Type))));
+            }
+            // Return the final TypeSyntax, aliased on not as required.
```
jskeet: Nit: on => or

Google.Api.Generator/RoslynUtils/Modifier.cs
```diff
@@ -0,0 +1,74 @@
...
+        Partial = 0x40,
+        Abstract = 0x80,
+        Virtual = 0x100,
+        Override = 0x200,
+        Internal = 0x400,
+    }
+
+    internal static class ModifierExtensions
+    {
+        private static readonly SyntaxToken PUBLIC_TOKEN = SyntaxFactory.Token(SyntaxKind.PublicKeyword);
```
jskeet: Any reason for shouty case here?

Google.Api.Generator/Utils/Disposable.cs
```diff
@@ -0,0 +1,25 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System;
+
+namespace Google.Api.Generator.Utils
+{
+    internal class Disposable : IDisposable
```
jskeet: I think a comment would be useful on this class, mostly to give an idea of when it's useful.

Google.Api.Generator/Utils/Typ.cs
```diff
@@ -0,0 +1,86 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Google.Api.Generator.Utils
+{
+    /// <summary>
+    /// Abstraction of a .NET type.
+    /// </summary>
+    internal abstract class Typ
```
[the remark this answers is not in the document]
chrisdunelm: The reason I chose `Typ` is because this is used extensively throughout the code and I wanted it to be very concise.
I agree it may not be immediately clear what it is, but the conciseness was more important.

`ClrType` I suppose may be more suitable, but it's over twice the number of characters!!

Google.Api.Generator/RoslynUtils/Modifier.cs
chrisdunelm: Not really - I can never remember the right format for these constant-like things ;)

Google.Api.Generator/Utils/Typ.cs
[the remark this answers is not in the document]
jskeet: Hmm. I definitely found it distracting to see Typ (and typ in comments) all over the place. It always made me think it was a mistake.
I guess you're going to be the one reading and writing this code the most, but it still makes me sad. I'd *at least* add a comment to explain the name oddity.

Google.Api.Generator/Generation/SourceFileContext.cs
```diff
@@ -0,0 +1,148 @@
...
+    {
+        public enum ImportStyle
+        {
+            FullyAliased,
+            // TODO: Unaliased
+        }
+
+        private static readonly IReadOnlyDictionary<string, string> s_wellknownNamespaceAliases = new Dictionary<string, string>
+        {
+            { typeof(System.Int32).Namespace, "sys" }, // Don't use "s"; one-letter alias cause a compilation error!
```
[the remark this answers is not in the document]
chrisdunelm: Thanks, done.

Google.Api.Generator/Generation/SourceFileContext.cs
chrisdunelm: done

Google.Api.Generator/RoslynUtils/Modifier.cs
chrisdunelm: Done

Google.Api.Generator/Utils/Disposable.cs
chrisdunelm: Done

Google.Api.Generator/Utils/Typ.cs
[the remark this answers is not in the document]
chrisdunelm: Done. Comment and TODO added.

PR #5, opened 2018-12-05
Add code whitespace formatter, with test

chrisdunelm:
Proto test can now use Roslyn to check generated code.

### DIFF
--- a/Google.Api.Generator.Tests/WhitespaceFormatterTest.cs
+++ b/Google.Api.Generator.Tests/WhitespaceFormatterTest.cs
@@ -15,13 +15,17 @@
 // This test loads this source file during the test, parses it into
 // Roslyn, removes all trivia, passes it through the Whitespace formatter,
 // then checks that the input source and re-formatted source are identical.
-// Code at the start, end, and between IGNORE_START/IGNORE_END markers
-// are ignored.
+// Code between the TEST_SOURCE_START and TEST_SOURCE_END comments are used
+// as source for this test.
 // So some of the source code looks a bit odd, as it needs to test every
 // code construct is formatted correctly.
 // TODO: Add further code as WhitespaceFormatter adds more formatting.
 
-// IGNORE_END
+// TODO: Consider splitting this test into multiple files; possibly
+// putting the test source separately, or splitting the test source
+// into multiple source files, and/or multiple tests.
+
+// TEST_SOURCE_START
 using Google.Api.Generator.Formatting;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
@@ -95,7 +99,7 @@ namespace Google.Api.Generator.Tests
         {
             // Test class that has no members.
         }
-        // IGNORE_START
+        // TEST_SOURCE_END
 
         private class TriviaRemover : CSharpSyntaxRewriter
         {
@@ -120,8 +124,8 @@ namespace Google.Api.Generator.Tests
             var testSourceLines = sourceLines.Aggregate((ignoring: true, lines: new List<string>()), (acc, line) =>
             {
                 var ignoring =
-                    line.Trim() == "// IGNORE_START" ? true :
-                    line.Trim() == "// IGNORE_END" ? false : (bool?)null;
+                    line.Trim() == "// TEST_SOURCE_END" ? true :
+                    line.Trim() == "// TEST_SOURCE_START" ? false : (bool?)null;
                 if (!acc.ignoring && ignoring != true && !line.Trim().StartsWith("//"))
                 {
                     acc.lines.Add(line);
@@ -143,8 +147,8 @@ namespace Google.Api.Generator.Tests
             // Check that the sources are identical.
             Assert.Equal(testSource, testSourceFormatter);
         }
-        // IGNORE_END
+        // TEST_SOURCE_START
     }
 }
 
-// IGNORE_START
+// TEST_SOURCE_END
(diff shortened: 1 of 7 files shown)


Google.Api.Generator.Tests/WhitespaceFormatterTest.cs
```diff
@@ -0,0 +1,150 @@
...
+// This test loads this source file during the test, parses it into
+// Roslyn, removes all trivia, passes it through the Whitespace formatter,
+// then checks that the input source and re-formatted source are identical.
+// Code at the start, end, and between IGNORE_START/IGNORE_END markers
+// are ignored.
+// So some of the source code looks a bit odd, as it needs to test every
+// code construct is formatted correctly.
+// TODO: Add further code as WhitespaceFormatter adds more formatting.
+
+// IGNORE_END
```
jskeet: This is odd. No IGNORE_START until later? Should I know what this means?

Google.Api.Generator.Tests/WhitespaceFormatterTest.cs
```diff
@@ -0,0 +1,150 @@
...
+                set => _p4 = value;
+            }
+        }
+
+        // Test base-list with one item.
+        private class EmptyInner : List<string>
+        {
+            // Test class that has no members.
+        }
+        // IGNORE_START
```
jskeet: Now we get the start... but I don't know what it means (yet).

Google.Api.Generator.Tests/WhitespaceFormatterTest.cs
```diff
@@ -0,0 +1,150 @@
...
+            var rootFormatter = new WhitespaceFormatter().Visit(rootWithoutWhitespace);
+            var testSourceFormatter = rootFormatter.ToFullString();
+            // Check that the sources are identical.
+            Assert.Equal(testSource, testSourceFormatter);
+        }
+        // IGNORE_END
+    }
+}
+
+// IGNORE_START
```
jskeet: Again, this is odd.

Google.Api.Generator.Tests/WhitespaceFormatterTest.cs
```diff
@@ -0,0 +1,150 @@
...
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Runtime.CompilerServices;
+using Xunit;
+using sys = System;
+
+namespace Google.Api.Generator.Tests
```
jskeet: Is there a significant benefit in making the test's own source code (with bits ignored) the test? While it's amusing/elegant, it feels like it may be making things trickier than expected.

Google.Api.Generator/RoslynUtils/RoslynExtensions.cs
```diff
@@ -0,0 +1,31 @@
...
+
+using Microsoft.CodeAnalysis;
+using System.Linq;
+using static Microsoft.CodeAnalysis.CSharp.SyntaxFactory;
+
+namespace Google.Api.Generator.RoslynUtils
+{
+    internal static class RoslynExtensions
+    {
+        public static T WithTrailingSpace<T>(this T node) where T : SyntaxNode => node.WithTrailingTrivia(Space);
```
jskeet: I've found the ordering and grouping of these to be odd. It's also a bit weird that we only need a generic WithLeadingCrlf, and only need a non-generic WithLeadingSpace, but I guess that's just the way it happens.

Google.Api.Generator/Formatting/WhitespaceFormatter.cs
```diff
@@ -0,0 +1,273 @@
...
+using Google.Api.Generator.Utils;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using static Microsoft.CodeAnalysis.CSharp.SyntaxFactory;
+
+namespace Google.Api.Generator.Formatting
```
jskeet: I haven't reviewed this in detail as the test is pretty compelling. Is there anything you'd like me to look at closely?

Google.Api.Generator.Tests/WhitespaceFormatterTest.cs
chrisdunelm: I've changed the markers to `TEST_SOURCE_START` and `TEST_SOURCE_END` which I think is somewhat clearer. The way the test is structured is explained in the comment above.

Google.Api.Generator.Tests/WhitespaceFormatterTest.cs
chrisdunelm: I think keeping this test within a single source file is useful.
I've added a TODO to reconsider this if splitting the test would be more comprehensible.

Google.Api.Generator.Tests/WhitespaceFormatterTest.cs
chrisdunelm: ack, see earlier response.

Google.Api.Generator.Tests/WhitespaceFormatterTest.cs
chrisdunelm: ack

Google.Api.Generator/Formatting/WhitespaceFormatter.cs
chrisdunelm: No, nothing specific here, it's all fairly simple/obvious code.

Google.Api.Generator/RoslynUtils/RoslynExtensions.cs
chrisdunelm: I don't see any particular ordering/grouping being particularly better than others unfortunately.
The methods here are the ones that are required so far.
Some of the "missing" methods will be added later once more generation code is added.

PR #6, opened 2018-12-06
Add first method to the Settings class

chrisdunelm:
Including extra code required to generate this method, and tests

### DIFF
--- a/Google.Api.Generator.Tests/ProtoTest.cs
+++ b/Google.Api.Generator.Tests/ProtoTest.cs
@@ -105,11 +105,11 @@ namespace Google.Api.Generator.Tests
             Assert.NotNull(ns);
             // Check there is one settings class.
             Assert.NotEmpty(ns.Members);
-            var settingsClss = ns.Members.OfType<ClassDeclarationSyntax>().Where(x => x.Identifier.Text == "TestSettings").ToList();
-            Assert.Single(settingsClss);
-            var settingsCls = settingsClss[0];
+            var settingsClasses = ns.Members.OfType<ClassDeclarationSyntax>().Where(x => x.Identifier.Text == "TestSettings").ToList();
+            Assert.Single(settingsClasses);
+            var settingsClass = settingsClasses[0];
             // Check the static method `GetDefault()` exists.
-            var getDefaultMethods = settingsCls.Members.OfType<MethodDeclarationSyntax>()
+            var getDefaultMethods = settingsClass.Members.OfType<MethodDeclarationSyntax>()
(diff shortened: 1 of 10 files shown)


Google.Api.Generator.Tests/ProtoTest.cs
```diff
@@ -105,8 +105,16 @@ public void TestThatSomeCodeIsGenerated()
             Assert.NotNull(ns);
             // Check there is one settings class.
             Assert.NotEmpty(ns.Members);
-            var settingsCls = ns.Members.Where(x => (x as ClassDeclarationSyntax)?.Identifier.Text == "TestSettings").ToList();
-            Assert.Single(settingsCls);
+            var settingsClss = ns.Members.OfType<ClassDeclarationSyntax>().Where(x => x.Identifier.Text == "TestSettings").ToList();
```
jskeet: Any reason not to use `settingsClasses` and `settingsClass` here? I feel this is brevity at the expense of readability.

Google.Api.Generator/Generation/ServiceSettingsCodeGenerator.cs
```diff
@@ -35,9 +36,21 @@ internal class ServiceSettingsCodeGenerator
 
         private ClassDeclarationSyntax Generate()
         {
-            var cls = Class(Public | Sealed | Partial, _svc.SettingsTyp, baseType: _ctx.Type<ServiceSettingsBase>());
-            // TODO: Fill the settings class.
+            var cls = Class(Public | Sealed | Partial, _svc.SettingsTyp, baseType: _ctx.Type<ServiceSettingsBase>())
+                .WithXmlDoc(XmlDoc.Summary("Settings for a ", _ctx.Type(_svc.ClientAbstractTyp), "."));
```
jskeet: Does the Java generator handle vowels properly, e.g. "an ImageAnnotatorClient" instead of "a ImageAnnotatorClient"? Maybe add a TODO?

Google.Api.Generator/Generation/SourceFileContext.cs
```diff
@@ -75,6 +75,13 @@ public IDisposable InNamespace(NamespaceDeclarationSyntax ns)
             return new Disposable(() => Namespace = restoreNamespace);
         }
 
+        public IDisposable InClass(ClassDeclarationSyntax cls)
```
jskeet: I'm having a hard time working out what this is trying to achieve. An XML comment would help. (Is it just a matter of saying "We're already in the given class, so we don't need to import it"?)

PR #7, opened 2018-12-06
Initial MethodDetails implementation

chrisdunelm:
Only supports 'normal' methods so far, still need to add paged, streaming, LRO, ...


Google.Api.Generator/Generation/MethodDetails.cs
```diff
@@ -0,0 +1,94 @@
+﻿using Google.Api.Gax.Grpc;
```
jskeet: Copyright? (I haven't checked other files - oops)

Google.Api.Generator/Utils/Typ.cs
```diff
@@ -56,6 +58,17 @@ private sealed class FromNested : Typ
             public override string FullName => $"{DeclaringTyp.FullName}+{Name}";
         }
 
+        private sealed class FromGeneric : Typ
+        {
+            public FromGeneric(Typ def, IEnumerable<Typ> typeArgs) => (_def, GenericArgTyps) = (def, typeArgs);
```
jskeet: Just to check, FromGeneric represents a constructed type, and def is the unbound generic type? It might be worth making a note of that somewhere.

Google.Api.Generator/Generation/MethodDetails.cs
chrisdunelm: Oops, yes, added.

Google.Api.Generator/Utils/Typ.cs
chrisdunelm: XmlDoc added.

PR #10, opened 2018-12-06
Add code required to generate settings copy ctor

chrisdunelm:
And add to the test.

### DIFF
--- a/Google.Api.Generator.Tests/ProtoTest.cs
+++ b/Google.Api.Generator.Tests/ProtoTest.cs
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
 using System;
@@ -85,10 +86,8 @@ namespace Google.Api.Generator.Tests
         [Fact]
         public void TestThatSomeCodeIsGenerated()
         {
-            // Test that the code generator executes, and produces *something*.
-            // At the moment the code produced is not formatted, which means all
-            // whitespace is missing, and is therefore unparsable as C# code.
-            // TODO: Remove this test.
+            // Test that the code generator executes, and produces the expected code.
+            // TODO: Improve this testing infrastructure.
             var files = Run("ProtoTest.proto", "testing");
             var clientCs = files.FirstOrDefault(x => x.RelativePath == "TestClient.cs");
             Assert.NotNull(clientCs);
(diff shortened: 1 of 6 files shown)


Google.Api.Generator.Tests/ProtoTest.cs
```diff
@@ -85,10 +86,8 @@ public void TestProtocExecution()
         [Fact]
         public void TestThatSomeCodeIsGenerated()
         {
-            // Test that the code generator executes, and produces *something*.
-            // At the moment the code produced is not formatted, which means all
-            // whitespace is missing, and is therefore unparsable as C# code.
-            // TODO: Remove this test.
+            // Test that the code generator executes, and produces the expected code.
+            // TODO: Improve this testing infrastructure.
```
jskeet: Yup, I think at this point we're getting to the stage where it would be good to have "expected output" to review. If you could parse that, normalize both the actual output and expected output, and compare them, that would be great.

Google.Api.Generator/RoslynUtils/RoslynBuilder.cs
```diff
@@ -57,6 +64,35 @@ public static ClassDeclarationSyntax Class(Modifier modifier, Typ typ, TypeSynta
...
+            {
+                accessors.Add(AccessorDeclaration(SyntaxKind.SetAccessorDeclaration).WithSemicolonToken(s_semicolonToken));
+            }
+            return PropertyDeclaration(type, name)
+                .AddModifiers(modifiers.ToSyntaxTokens())
+                .AddAccessorListAccessors(accessors.ToArray());
+        }
+
+        public static ExpressionSyntax Nameof(ParameterSyntax nameof) =>
+            InvocationExpression(IdentifierName("nameof"))
```
jskeet: There's a certain irony in requiring a string literal here, in code to avoid string literals.

Google.Api.Generator/RoslynUtils/RoslynConverters.cs
```diff
@@ -86,5 +87,36 @@ public static IEnumerable<StatementSyntax> ToStatements(object o)
...
+        {
+            if (o.GetType().IsEnum)
+            {
+                if (genericArgs.Any())
+                {
+                    throw new ArgumentException("Generic args must not be present for an enum");
+                }
+                return IdentifierName(o.ToString());
+            }
+            SyntaxToken name;
```
jskeet: Looking forward to switch expressions much? ;)

Google.Api.Generator.Tests/ProtoTest.cs
chrisdunelm: Completely agree. I'll work on the testing in a PR soon.
(At the moment, I think PR means "Pouring with Rain")

Google.Api.Generator/RoslynUtils/RoslynBuilder.cs
chrisdunelm: Indeed ;)
But at least this helper method means it's only here once :)

Google.Api.Generator/RoslynUtils/RoslynConverters.cs
chrisdunelm: Definitely.
I've been waiting for switch expressions for many C# versions! Scala has always had them, and they're lovely :)
I vote we switch to VS2019 and C# 8 previews *now*!

PR #12, opened 2018-12-11
Add LRO settings

chrisdunelm:
The proto-generated C# code in the new `Google.Api.CommonProtosExtra` don't need reviewing; they are protos in the `input-contract` branch of `api-common-protos` that aren't yet in GAX.


Google.Api.Generator.Tests/ProtoTest.cs
```diff
@@ -83,29 +84,25 @@ public void ProtocExecution()
             Run("ProtoTest.proto", "testing");
         }
 
-        [Fact]
-        public void ProtoTests()
+        private void ProtoTestSingle(string testProtoName)
```
jskeet: Rather than have separate tests, use a Theory here with InlineData?

Google.Api.Generator/Generation/ServiceSettingsCodeGenerator.cs
```diff
@@ -89,14 +96,40 @@ private IEnumerable<PropertyDeclarationSyntax> SettingsProperties()
...
+                {
+                    // TODO: Extra properties for streaming methods.
+                    case MethodDetails.Lro lro:
+                        yield return LroSettingsProperty(lro);
+                        break;
+                }
             }
         }
 
+        private PropertyDeclarationSyntax LroSettingsProperty(MethodDetails.Lro method) =>
```
jskeet: Yoicks! (No ideas about how to simplify this, but eek!)

Google.Api.Generator/RoslynUtils/RoslynExtensions.cs
```diff
@@ -61,6 +64,22 @@ private static T WithBody<T>(IEnumerable<object> code, Func<ArrowExpressionClaus
         public static MethodDeclarationSyntax WithBody(this MethodDeclarationSyntax method, params object[] code) =>
             WithBody(code, x => method.WithExpressionBody(x).WithSemicolonToken(s_semicolonToken), method.WithBody);
 
+        public static ObjectCreationExpressionSyntax WithInitializer(
+            this ObjectCreationExpressionSyntax obj, params (string propertyName, object code)[] inits)
+        {
+            if (!obj.ArgumentList.Arguments.Any())
+            {
+                // If calling the parameterless ctor, recreate the `new` expression without an argument list.
```
jskeet: Thank you! (I assume this means `new Foo { Bar = X }` rather than `new Foo() { Bar = X }`?)

Google.Api.Generator.Tests/ProtoTest.cs
chrisdunelm: I'm using separate tests as it makes it much easier to execute just a single test case; which is very useful during this development.
I'll add a TODO in the next PR to consider changing it later.

Google.Api.Generator/Generation/ServiceSettingsCodeGenerator.cs
chrisdunelm: Indeed. This is a downside of not using templates.

Google.Api.Generator/RoslynUtils/RoslynExtensions.cs
chrisdunelm: Yes, exactly.

PR #13, opened 2018-12-11
Allow partial expected files in generator tests

chrisdunelm:


chrisdunelm:
I should have provided more information in the PR description...

There is a change to the test, which means that only the expected lines between `// TEST_START` and `// TEST_END` are considered. So long as these lines are exactly present in the generated code then the test passes.
If there are no `// TEST_...` comments then the entire file needs to match.

This is useful because there will be many test cases coming soon where I definitely don't want to check the *entire* output is correct, just the one feature that is being tested. This will also become more important as the generated output increases in size.
So in this PR, for example, the LRO test now only checks that the LRO-specific code is present, as that test-cases is only for testing LRO.

PR #14, opened 2018-12-11
Add retry settings

chrisdunelm:
Retry has changed from the current config/generator. There are various ways we can mantain compatibility which we can discuss later.

Please excuse the crazy long lines, I'll get a line-length-formatter in soon ;)

### DIFF
--- a/Google.Api.Generator/Formatting/WhitespaceFormatter.cs
+++ b/Google.Api.Generator/Formatting/WhitespaceFormatter.cs
@@ -151,6 +151,13 @@ namespace Google.Api.Generator.Formatting
             return node;
         }
 
+        public override SyntaxNode VisitNameColon(NameColonSyntax node)
+        {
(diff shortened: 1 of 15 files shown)


Google.Api.Generator/Formatting/WhitespaceFormatter.cs
```diff
@@ -151,6 +151,13 @@ public override SyntaxNode VisitArgumentList(ArgumentListSyntax node)
             return node;
         }
 
+        public override SyntaxNode VisitNameColon(NameColonSyntax node)
```
jskeet: Is this syntax used for inheritance as well? Just wondering whether it's going to cause us problems when we want a space before the colon for inheritance, but not for named arguments.

chrisdunelm: No, inheritance uses `VisitBaseList`, which includes the `:` prefix of the base list.

PR #15, opened 2019-01-17
Client abstract class with just default host and scope properties

chrisdunelm:


Google.Api.Generator.Tests/ProtoTest.cs
```diff
@@ -126,9 +126,11 @@ private void ProtoTestSingle(string testProtoName)
         }
 
         // TODO: Consider using a single Theory for these.
+        // It's currently convinient having them separate, so they can easily be run individually.
```
jskeet: Nit: convenient

chrisdunelm: Oops, done.

PR #18, opened 2019-01-18
Abstract basic RPC methods

chrisdunelm:


Google.Api.Generator.Tests/ProtoTests/BasicClient.cs
```diff
@@ -119,5 +120,41 @@ public static BasicClient Create(grpccore::CallInvoker callInvoker, BasicSetting
 
         /// <summary>The underlying gRPC Basic client</summary>
         public virtual Basic.BasicClient GrpcClient => throw new sys::NotImplementedException();
+
+        /// <summary>Test summary text for IdempotentMethod</summary>
+        /// <param name="request">The request object containing all of the parameters for the API call.</param>
+        /// <param name="callSettings">If not null, applies overrides to this RPC call.</param>
+        /// <returns>The RPC response.</returns>
+        public virtual Response IdempotentMethod(Request request, gaxgrpc::CallSettings callSettings = null) => throw new sys::NotImplementedException();
```
jskeet: How hard would it be to add a line break after `=>`? Definitely not critical, but would reduce the number of places we have huge lines. Definitely one to do later rather than now.

Google.Api.Generator/Generation/ServiceMethodGenerator.cs
```diff
@@ -0,0 +1,106 @@
...
+                    .WithXmlDoc(SummaryXmlDoc, RequestXmlDoc, CallSettingsXmlDoc, ReturnsSyncXmlDoc);
+
+            public MethodDeclarationSyntax AbstractAsyncCallSettingsRequestMethod =>
+                Method(Public | Virtual, Ctx.Type(MethodDetails.AsyncReturnTyp), MethodDetails.AsyncMethodName)(RequestParam, CallSettingsParam)
+                    .WithBody(Throw(New(Ctx.Type<NotImplementedException>())()))
+                    .WithXmlDoc(SummaryXmlDoc, RequestXmlDoc, CallSettingsXmlDoc, ReturnsAsyncXmlDoc);
+
+            public MethodDeclarationSyntax AbstractAsyncCancellationTokenRequestMethod =>
+                Method(Public | Virtual, Ctx.Type(MethodDetails.AsyncReturnTyp), MethodDetails.AsyncMethodName)(RequestParam, CancellationTokenParam)
+                        .WithBody(This.Call(AbstractAsyncCallSettingsRequestMethod)(RequestParam, Ctx.Type<CallSettings>().Call(nameof(CallSettings.FromCancellationToken))(CancellationTokenParam)))
```
jskeet: Outdent this line and the line below?

Google.Api.Generator.Tests/ProtoTests/BasicClient.cs
chrisdunelm: Yes, formatting will be done in a later PR, and will use a "proper" line-length-aware formatter that'll make everything look beautiful.
Fortunately that's an orthogonal problem to the code-gen, so can be considered independently.

Google.Api.Generator/Generation/ServiceMethodGenerator.cs
chrisdunelm: Oops, yes, will fix.

PR #20, opened 2019-01-18
Service implementation initialization code

chrisdunelm:


Google.Api.Generator.Tests/ProtoTests/BasicClient.cs
```diff
@@ -161,11 +162,36 @@ public static BasicClient Create(grpccore::CallInvoker callInvoker, BasicSetting
     /// <summary>Basic client wrapper implementation, for convenient use.</summary>
     public sealed partial class BasicClientImpl : BasicClient
     {
+        private readonly gaxgrpc::ApiCall<Request, Response> _callIdempotentMethod;
+
```
jskeet: When we get round to formatting, let's not have lines between all the field declarations. Not for now though.

PR #21, opened 2019-01-18
Implementation of RPC methods.

chrisdunelm:


chrisdunelm:
They are in the abstract class and forwarded to the `CallSettings` method; so not required in the implementation class.

jskeet:
Doh - I should really concentrate harder...

chrisdunelm:
:)

PR #22, opened 2019-01-18
LRO abstract methods; test refactor

chrisdunelm:


Google.Api.Generator.Tests/ProtoTests/ProtoMsgFake.cs
```diff
@@ -0,0 +1,19 @@
+﻿using Google.Protobuf;
+using Google.Protobuf.Reflection;
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Testing
+{
+    public abstract class ProtoMsgFake<T> : IMessage<T> where T : ProtoMsgFake<T>
```
jskeet: I think this may be the only non-generated implementation of `IMessage<T>` I've ever seen...

Google.Api.Generator.Tests/ProtoTests/BasicLro/BasicLroClient.cs
```diff
@@ -0,0 +1,74 @@
...
+        {
+            DefaultPollSettings = new gax::PollSettings(gax::Expiration.FromTimeout(sys::TimeSpan.FromHours(24)), sys::TimeSpan.FromSeconds(20), 1.5, sys::TimeSpan.FromSeconds(45)),
+        };
+        // TEST_END
+    }
+
+    public abstract class BasicLroClient
+    {
+        // TEST_START
+        /// <summary></summary>
```
jskeet: Any reason comments aren't in here yet?

Google.Api.Generator.Tests/ProtoTests/BasicLro/BasicLroClient.cs
```diff
@@ -0,0 +1,74 @@
...
+using gaxgrpc = Google.Api.Gax.Grpc;
+using lro = Google.LongRunning;
+using sys = System;
+using st = System.Threading;
+using stt = System.Threading.Tasks;
+
+namespace Testing.Basiclro
+{
+    /// <summary>Settings for a <see cref="BasicLroClient"/>.</summary>
+    public sealed partial class BasicLroSettings : gaxgrpc::ServiceSettingsBase
```
jskeet: I haven't checked - I assume this basically looks like the code we currently generate? (For both settings and the client.)
Anything you'd like me to look at carefully?

Google.Api.Generator.Tests/ProtoTests/BasicLro/BasicLroClient.cs
chrisdunelm: It's because there isn't a comment in the proto; so it's correct that the summary is empty here.
The "Basic" test tests that comments work, so I don't think there's a reason to also have them in this test.

Google.Api.Generator.Tests/ProtoTests/BasicLro/BasicLroClient.cs
chrisdunelm: Yes, it's all the same :)

Google.Api.Generator.Tests/ProtoTests/ProtoMsgFake.cs
chrisdunelm: :)

PR #25, opened 2019-01-24
Improved XmlDoc formatting

chrisdunelm:


Google.Api.Generator.Tests/ProtoTests/BasicLro/BasicLroClient.cs
```diff
@@ -37,35 +48,56 @@ private BasicLroSettings(BasicLroSettings existing)
...
         /// <param name="callSettings">If not null, applies overrides to this RPC call.</param>
         /// <returns>A Task containing the RPC response.</returns>
         public virtual stt::Task<lro::Operation<LroResponse, LroMetadata>> Method1Async(Request request, gaxgrpc::CallSettings callSettings = null) => throw new sys::NotImplementedException();
 
-        /// <summary></summary>
-        /// <param name="request">The request object containing all of the parameters for the API call.</param>
-        /// <param name="cancellationToken">A <see cref="st::CancellationToken"/> to use for this RPC.</param>
+        /// <summary>
+        /// </summary>
+        /// <param name="request">
```
jskeet: How hard would it be to make these "single line" items like `<returns>`? I think that would usually be a bit nicer.

chrisdunelm: It's based on line-length. The XmlDoc formatter is currently using 100 characters as the maximum length. It'll single-line `param` xmldoc if it'll fit within this limit; otherwise it splits it only multiple lines like this.
If you think 100 characters is to conservative, this is trivial to increase ;) I was considering using 120.

jskeet: Interesting. Yes, I think 120 would be fine.

PR #32, opened 2019-01-27
Method Signatures (flattening) of basic RPCs

chrisdunelm:
Method Signatures within Paginated, LRO, etc... RPCs coming in a future PR

### DIFF
--- a/Google.Api.Generator/ProtoUtils/ProtoExtensions.cs
+++ b/Google.Api.Generator/ProtoUtils/ProtoExtensions.cs
@@ -12,7 +12,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+using Google.Protobuf;
 using Google.Protobuf.Reflection;
+using System;
+using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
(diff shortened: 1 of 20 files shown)


Google.Api.Generator/ProtoUtils/ProtoExtensions.cs
```diff
@@ -50,8 +53,55 @@ public static string CSharpNamespace(this FileDescriptor desc)
...
                 acc => acc.sb.ToString());
 
-        public static string CSharpName(this FieldDescriptor field) => CSharpName(field.Name, isFieldName: false);
+        public static string CSharpPropertyName(this FieldDescriptor field) => CSharpName(field.Name, isFieldName: false);
 
         public static string CSharpFieldName(this FieldDescriptor field) => CSharpName(field.Name, isFieldName: true);
+
+        private static IEnumerable<(ulong number, ByteString byteString)> GetRepeated(CustomOptions opts, int field)
+        {
+            // There is no way of reading a repeated custom option.
```
jskeet: I seem to remember thinking they never happened :) Have you filed a feature request? It may not be too hard to implement.

chrisdunelm: That didn't occur to me ;) Yes, I'll file a request.

PR #34, opened 2019-02-07
Initial resource-name code

chrisdunelm:
There are many different scenarios in which resource-names may be used. Only the basics are tested in this PR. More tests will be added in future PRs.


Google.Api.Generator.Tests/ProtoTests/ResourceNames/ResourceNamesResourceNames.cs
```diff
@@ -0,0 +1,99 @@
+﻿using gax = Google.Api.Gax;
```
jskeet: No copyright - I may just not have noticed this from previous files. Is that still to-do in general?

Google.Api.Generator.Tests/ProtoTests/ResourceNames/ResourceNamesResourceNames.cs
```diff
@@ -0,0 +1,99 @@
...
+        /// <inheritdoc/>
+        public static bool operator ==(SimpleResourceName a, SimpleResourceName b) => ReferenceEquals(a, b) || (a?.Equals(b) ?? false);
+
+        /// <inheritdoc/>
+        public static bool operator !=(SimpleResourceName a, SimpleResourceName b) => !(a == b);
+    }
+
+    public partial class SimpleResource
+    {
+        /// <summary>
```
jskeet: I can't remember what determines whether we do this or just a SimpleResourceName property. Possibly worth just checking, but I suspect you're right :)

Google.Api.Generator/Generation/ResourceNamesGenerator.cs
```diff
@@ -0,0 +1,261 @@
...
+                OperatorDeclarationSyntax InequalityOperator()
+                {
+                    var a = Parameter(_ctx.Type(def.One.ResourceNameTyp), "a");
+                    var b = Parameter(_ctx.Type(def.One.ResourceNameTyp), "b");
+                    return OperatorMethod(_ctx.Type<bool>(), "!=")(a, b)
+                        .WithBody(Not(Parens(a.Equality(b))))
+                        .WithXmlDoc(XmlDoc.InheritDoc);
+                }
+            }
+
```
jskeet: Random blank line?

chrisdunelm: Oops, yes, will fix.

PR #36, opened 2019-02-08
Fix generator test

chrisdunelm:
It used to allow extra lines in the generated output, which were not present in the expected code. Now it only accepts generated code that is exactly as in the expected code blocks.

### DIFF
--- a/Google.Api.Generator.Tests/ProtoTest.cs
+++ b/Google.Api.Generator.Tests/ProtoTest.cs
@@ -96,31 +96,66 @@ namespace Google.Api.Generator.Tests
                 Assert.True(File.Exists(expectedFilePath), $"Expected file does not exist: '{expectedFilePath}'");
                 var expectedLines = File.ReadAllLines(expectedFilePath).Select(x => x.Trim('\r')).ToList();
                 var actualLines = Encoding.UTF8.GetString(file.Content).Split('\n').Select(x => x.Trim('\r')).ToList();
-                // Check that all the testable lines with in the expected file match.
-                int expectedIndex = 0;
-                int actualIndex = 0;
-                bool active = !expectedLines.Any(x => x.Trim() == "// TEST_START");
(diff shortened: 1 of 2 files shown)


Google.Api.Generator.Tests/ProtoTest.cs
```diff
@@ -96,31 +96,66 @@ private void ProtoTestSingle(string testProtoName)
...
+                            }
+                            blockIndex = 0;
+                        }
                     }
-                    else
+                    if (missing.line != null)
                     {
-                        active |= expectedLines[expectedIndex].Trim() == "// TEST_START";
-                        expectedIndex += 1;
+                        Console.WriteLine(string.Join(Environment.NewLine, actualLines));
```
jskeet: Is this deliberately still in here?

Google.Api.Generator.Tests/ProtoTest.cs
```diff
@@ -96,31 +96,66 @@ private void ProtoTestSingle(string testProtoName)
...
+                var expectedBlocks = expectedLines.Any(x => x.Trim() == "// TEST_START") ? TestBlocks() : new[] { expectedLines.Select((s, i) => (i, s)).ToList() };
+                // Check that all expected code blocks in the expected code exist in the generated (actual) code.
+                // The order of the test blocks is not enforced, only the content.
+                foreach (var block in expectedBlocks)
                 {
-                    active &= expectedLines[expectedIndex].Trim() != "// TEST_END";
-                    if (active)
+                    int blockIndex = 0;
+                    (int lineNumber, string line) missing = default;
+                    int missingBestLength = -1;
```
jskeet: I don't really understand this code, to be honest. What's the "best length" here?

chrisdunelm:
I'll tidy/refactor/comment this in a future PR...
Yes, the `Console.WriteLine` is meant to be there. It prints out the generated output on test failure. Very useful for debugging :)

PR #37, opened 2019-02-08
Fix naming errors

chrisdunelm:
Fix casing of resource-name class Parse() method parameter

Fix name of reosurce-name property when the underlying property is named 'Name'

Improve test output when proto test fails

### DIFF
--- a/Google.Api.Generator/Utils/SystemExtensions.cs
+++ b/Google.Api.Generator/Utils/SystemExtensions.cs
@@ -20,8 +20,9 @@ namespace Google.Api.Generator.Utils
     internal static class SystemExtensions
     {
         private static string Camelizer(string s, bool firstUpper) =>
-            s.Aggregate((upper: firstUpper, sb: new StringBuilder()), (acc, c) =>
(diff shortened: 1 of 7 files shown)


Google.Api.Generator/Utils/SystemExtensions.cs
```diff
@@ -20,8 +20,9 @@ namespace Google.Api.Generator.Utils
     internal static class SystemExtensions
     {
         private static string Camelizer(string s, bool firstUpper) =>
-            s.Aggregate((upper: firstUpper, sb: new StringBuilder()), (acc, c) =>
-                c == '_' ? (true, acc.sb) : (false, acc.sb.Append(acc.upper ? char.ToUpperInvariant(c) : char.ToLowerInvariant(c))),
+            s.Aggregate((upper: (bool?)firstUpper, sb: new StringBuilder()), (acc, c) =>
+                c == '_' ? (true, acc.sb) :
+                    ((bool?)null, acc.sb.Append(acc.upper is bool upper ? upper ? char.ToUpperInvariant(c) : char.ToLowerInvariant(c) : c)),
```
jskeet: I think I'd probably extract the last part into a separate MaybeForceCase method or similar. It took me a while to work out what on earth was going on.

chrisdunelm: Done

PR #43, opened 2019-02-12
Support for wildcard resources

chrisdunelm:


Google.Api.Generator.Tests/ProtoTests/ResourceNames/ResourceNamesClient.cs
```diff
@@ -254,6 +254,155 @@ public abstract class ResourceNamesClient
...
+        public virtual Response WildcardMethod(string name, scg::IEnumerable<string> names, string requiredSingle, scg::IEnumerable<string> requiredRepeated, gaxgrpc::CallSettings callSettings = null) =>
+            WildcardMethod(new WildcardResource
+            {
+                Name = name ?? "",
+                Names =
+                {
+                    names ?? linq::Enumerable.Empty<string>(),
+                },
+                RequiredSingle = gax::GaxPreconditions.CheckNotNullOrEmpty(requiredSingle, nameof(requiredSingle)),
+                RequiredRepeated =
```
jskeet: I'm always somewhat dubious of "required repeated" - it's not like we're checking that it's non-empty. Probably worth further consideration.

Google.Api.Generator/Generation/ResourceNamesGenerator.cs
```diff
@@ -235,15 +237,14 @@ private IEnumerable<ClassDeclarationSyntax> ProtoMessagePartials()
                                 if (res.field.IsRepeated)
                                 {
                                     var repeatedTyp = Typ.Generic(typeof(ResourceNameList<>), one.ResourceNameTyp);
+                                    var s = Parameter(null, "s");
```
jskeet: Move this into the IsWildcard scope? I don't think it's needed outside it.

Google.Api.Generator.Tests/ProtoTests/ResourceNames/ResourceNamesClient.cs
chrisdunelm: Yes, I agree, and we may want to discuss changing how we support this.
However, this matches the existing code-gen, so is "correct".

Google.Api.Generator/Generation/ResourceNamesGenerator.cs
chrisdunelm: `s` is also used in the `else` clause. The unhelpfully short name doesn't help.

PR #52, opened 2019-02-20
Signature and snippet support for enums

chrisdunelm:


Google.Api.Generator/ProtoUtils/ProtoExtensions.cs
```diff
@@ -53,6 +53,16 @@ public static string CSharpNamespace(this FileDescriptor desc)
 
         public static string CSharpFieldName(this FieldDescriptor field) => field.Name.ToLowerCamelCase();
 
+        public static string CSharpName(this EnumValueDescriptor desc)
```
jskeet: Sorry for the duplicate code here :(
It might be worth linking to the C++ for reference. For example, link to https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/compiler/csharp/csharp_helpers.cc and suggest finding GetEnumValueName.

(I suspect this code may need to be adjusted to be a closer copy of that, too.)

chrisdunelm: Thanks for the reference. That prefix removal is a little more complex than I expected! I'll update the code here.

Google.Api.Generator.Tests/ProtoExtensionsTest.cs
```diff
@@ -0,0 +1,33 @@
...
+        [Theory]
+        [InlineData("enum", "not_the_same", "NotTheSame")]
+        [InlineData("enum", "enum", "Enum")]
+        [InlineData("enum", "enum_2", "_2")]
+        [InlineData("enum_type", "enumtype_value", "Value")]
+        [InlineData("enumtype", "enum_type_value", "Value")]
+        [InlineData("enum_type", "ENUMTYPEVALUE", "Value")]
+        [InlineData("enum_type", "ENUMTYPE_VALUE_2", "Value2")]
+        public void RemoveEnumPrefix(string enumName, string valueName, string expectedCSharpName) =>
+            Assert.Equal(expectedCSharpName, ProtoExtensions.RemoveEnumPrefix(enumName, valueName));
```
jskeet: Looks like you have some tests I don't (`enum_type_value`), but you could potentially port everything from here too:
https://github.com/protocolbuffers/protobuf/blob/d9ccd0c0e6bbda9bf4476088eeb46b02d7dcd327/src/google/protobuf/compiler/csharp/csharp_generator_unittest.cc

chrisdunelm: Thanks, added.

PR #53, opened 2019-02-20
Pure refactor of snippet generation

chrisdunelm:
In preparation of snippet generation of method signatures.


chrisdunelm:
One of the many beauties of having tests :)

PR #57, opened 2019-02-22
Snippet support for LRO signatures and resource-names

chrisdunelm:


jskeet:
At a conference today, so it'll be Monday probably.

chrisdunelm:
Np :) Happy conferencing.
I'll merge now, but very happy to change things in a future PR.

PR #63, opened 2019-02-25
Add missing paging parameters to paginated signature methods

chrisdunelm:


Google.Api.Generator/RoslynUtils/XmlDoc.cs
```diff
@@ -42,6 +42,8 @@ private static XmlNodeSyntax ToNode(object o)
                     return C("true");
                 case false:
                     return C("false");
+                case int v:
+                    return C(v.ToString());
```
jskeet: Possibly pass in the invariant culture here.

chrisdunelm: Ooh yes, will do in a future PR - just merged by accident!

PR #66, opened 2019-02-25
Async snippets for paginated methods

chrisdunelm:
Resource-names still todo.


chrisdunelm:
Pushed a change because travis is timing-out when running `dotnet`.

PR #69, opened 2019-02-26
Routing headers implementation, and test

chrisdunelm:


chrisdunelm:
Travis error is due to a problem installing the dotnet SDK. Will try it again tomorrow.

chrisdunelm:
Travis now working again :)

Google.Api.Generator/Generation/MethodDetails.cs
```diff
@@ -232,13 +241,79 @@ private MethodDetails(ServiceDetails svc, MethodDescriptor desc)
...
+                        if (finalField.FieldType != FieldType.String)
+                        {
+                            throw new InvalidOperationException($"Path in http url must resolve to a string field: '{path}'.");
+                        }
+                        yield return new RoutingHeader(WebUtility.UrlEncode(path), propertyNames);
+                    }
+                }
+            }
+
+            IEnumerable<string> ExtractBracedPaths(string s)
```
jskeet: I think you can use https://github.com/googleapis/gax-dotnet/blob/master/Google.Api.Gax/PathTemplate.cs instead.

chrisdunelm: `PathTemplate` doesn't support the `{name=some.thing}` format. Trying to use it throws here: https://github.com/googleapis/gax-dotnet/blob/7dac7db0502d55eae3e9778d22d08e6112f5639b/Google.Api.Gax/PathTemplate.cs#L435

jskeet: Ah, interesting. Maybe we should fix that then. (Not that we use PathTemplate much, as far as I'm aware. But it is *meant* to support that annotation...)

PR #71, opened 2019-02-27
Start of unit test generation

chrisdunelm:
Only generates sync request-method generation at the moment; and is not yet well-tested.

### DIFF
--- a/Google.Api.Generator.Tests/ProtoTest.cs
+++ b/Google.Api.Generator.Tests/ProtoTest.cs
@@ -94,9 +94,9 @@ namespace Google.Api.Generator.Tests
             // Verify each output file.
             foreach (var file in files)
             {
-                if (ignoreCsProj && file.RelativePath.EndsWith(".csproj") ||
-                    ignoreSnippets && file.RelativePath.Contains($".Snippets{Path.DirectorySeparatorChar}") ||
-                    ignoreUnitTests && file.RelativePath.Contains($".Tests{Path.DirectorySeparatorChar}"))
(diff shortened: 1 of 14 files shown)


Google.Api.Generator.Tests/ProtoTest.cs
```diff
@@ -94,11 +94,9 @@ private void ProtoTestSingle(string testProtoName, bool ignoreCsProj = false, bo
             // Verify each output file.
             foreach (var file in files)
             {
-                if (ignoreCsProj && file.RelativePath.EndsWith(".csproj"))
-                {
-                    continue;
-                }
-                if (ignoreSnippets && file.RelativePath.Contains($".Snippets{Path.DirectorySeparatorChar}"))
+                if (ignoreCsProj && file.RelativePath.EndsWith(".csproj") ||
```
jskeet: I honestly can't remember precedence rules for `&&` and `||`, but I think the code would at least be clearer with this being explicit:

```csharp
if ((ignoreCsProj && file.RelativePath.EndsWith(".csproj")) ||
    (... && ...) ||
    (... && ...))
```

chrisdunelm: Oops, yes, extra parentheses added.

chrisdunelm:
PTAL

Google.Api.Generator.Tests/ProtoTest.cs
chrisdunelm: And something you can't remember about C#! Incredible!!

jskeet:
(Was interrupted when reviewing, hence single comment. More later.)

PR #73, opened 2019-02-27
Add test for unit tests; test test-value generator

chrisdunelm:


Google.Api.Generator.Tests/ProtoTests/UnitTests/Testing.UnitTests.Tests/UnitTestsClientTest.g.cs
```diff
@@ -0,0 +1,108 @@
...
+namespace Testing.UnitTests
+{
+    /// <summary>Generated unit tests.</summary>
+    public sealed class GeneratedUnitTestsTest
+    {
+        [xunit::FactAttribute]
+        public void MethodValuesRequestObject()
+        {
+            moq::Mock<UnitTests.UnitTestsClient> mockGrpcClient = new moq::Mock<UnitTests.UnitTestsClient>(moq::MockBehavior.Strict);
+            ValuesRequest request = new ValuesRequest
```
jskeet: I'm not sure I see the value of setting all the properties here. The mock's just validating that it's receiving an equal object, and in fact it will be a reference to the *same* object, right?

Google.Api.Generator.Tests/ProtoTests/UnitTests/Testing.UnitTests/UnitTestsClient.cs
```diff
@@ -0,0 +1,272 @@
+﻿using gax = Google.Api.Gax;
```
jskeet: Do we need this file, as opposed to skipping it? Maybe I've missed something about how all of this infrastructure works.

Google.Api.Generator.Tests/ProtoTests/UnitTests/UnitTests.proto
```diff
@@ -0,0 +1,84 @@
...
+};
+
+message AnotherMessage {
+}
+
+enum Enum {
+  DEFAULT = 0; // First value must be 0 and is the default
+}
+
+message ValuesRequest {
```
jskeet: Note that this doesn't include a oneof with two fields. I haven't looked terribly closely at the generation code, but my guess is that if it *did* include a oneof, we'd end up generating a test with an odd proto.

Google.Api.Generator.Tests/ProtoTests/UnitTests/Testing.UnitTests/UnitTestsClient.cs
chrisdunelm: This file is required to allow the generated tests to execute.
Because the "expected" generated test files are present in the test project, they execute within the "dotnet test" run; which is nice, as it means we always verify that the generated tests (= expected test code) pass.

Google.Api.Generator.Tests/ProtoTests/UnitTests/UnitTests.proto
chrisdunelm: Right, yes, thanks.
I'll put a TODO in now, and add proper one-of support in a future PR.
The current (GAPIC monolith) test generation also doesn't handle one-ofs properly.

Google.Api.Generator.Tests/ProtoTests/UnitTests/Testing.UnitTests.Tests/UnitTestsClientTest.g.cs
chrisdunelm: Yes, you're right.
However, this is the "one true way" of writing these tests. And I suppose is more helpful if someone is looking at the tests to learn from them. And I feel it's probably the best thing to do, even though it's pointless. Like this entire test.

PR #74, opened 2019-02-27
Generate unit-test csproj; and add to tests

chrisdunelm:


Google.Api.Generator.Tests/ProtoTests/Basic/Testing.Basic.Tests/Testing.Basic.Tests.csproj
```diff
@@ -0,0 +1,19 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>netcoreapp2.2;net452</TargetFrameworks>
```
jskeet: For the generated tests, maybe only test netcoreapp? That will give us a simpler project, and the tests will be simpler to run too.

If this is currently for parity, that's fine - maybe add a TODO.

chrisdunelm: Good idea. I'll change it now.

PR #76, opened 2019-02-27
Generate async request method tests

chrisdunelm:


Google.Api.Generator.Tests/ProtoTests/Basic/Testing.Basic.Tests/BasicClientTest.g.cs
```diff
@@ -20,6 +21,19 @@ public void IdempotentMethodRequestObject()
             mockGrpcClient.VerifyAll();
         }
 
+        [xunit::FactAttribute]
```
jskeet: We have two overloads, one accepting a CancellationToken and one accepting CallSettings. Are we happy enough to only test one of these?

chrisdunelm: I'll add a test on both, for completenesses sake.

PR #78, opened 2019-02-28
Support unit-testing on services that contain LRO clients

chrisdunelm:
This does not add support for testing LRO methods; it just allows tests within services that *contain* LRO methods to work properly.

Unit-test generation is now at parity with the existing generator


chrisdunelm:
Ah, no, not quite at parity yet...
Still need to add unit tests for method signatures.

PR #88, opened 2019-03-06
Refactor tests to use Invoker, and not require file copying

chrisdunelm:


Google.Api.Generator.Tests/Invoker.cs
```diff
@@ -94,10 +102,12 @@ private static void Execute(string executable, string args, string workingDirect
...
         }
 
-        public static void Protoc(string args) => Execute(ProtocFile, args, null, TimeSpan.FromSeconds(15), 0);
+        // Generous timeouts, as travis can be slow sometimes, and many invocations may be executing concurrently.
+
+        public static void Protoc(string args) => Execute(ProtocFile, args, null, TimeSpan.FromSeconds(30), 0);
 
         public static void Dotnet(string args, string workingDirectory = null, int exitCode = 0) =>
-            Execute("dotnet", args, workingDirectory ?? GeneratorDir, TimeSpan.FromSeconds(90), exitCode);
+            Execute("dotnet", args, workingDirectory ?? GeneratorDir, TimeSpan.FromSeconds(120), exitCode);
```
jskeet: Does that suggest we've got some tests that take more than 90 seconds?

PR #91, opened 2019-03-13
Clone LRO operations settings in settings copy ctor

chrisdunelm:


chrisdunelm:
Yes, this is fixing a bug in the micro-generator.

PR #95, opened 2019-03-14
Add license to generated files

chrisdunelm:


Google.Api.Generator/Generation/SourceFileContext.cs
```diff
@@ -301,5 +301,50 @@ public ArrayTypeSyntax ArrayType(Typ arrayTyp, int? size = null)
         public virtual T Import<T>(System.Type t, T o) => o;
 
         public abstract CompilationUnitSyntax CreateCompilationUnit(NamespaceDeclarationSyntax ns);
+
+        protected CompilationUnitSyntax AddLicense(CompilationUnitSyntax cu)
+        {
+            // TODO: POssibly allow customization of license? Is this required?
+            // TODO: How should the copyright year be dealt with. would it be correct to always use the current year?
```
jskeet: That's what we do now, I believe - and it's probably a good enough starting point.

chrisdunelm:
The travis failure is almost certainly due to line-ending differences on Linux/Windows. I thought I'd allowed for that, but maybe not!
I'll fix soon.

Google.Api.Generator/Generation/SourceFileContext.cs
chrisdunelm: OK, thanks.

PR #100, opened 2019-04-04
[DO NOT MERGE] resource to resource-set change, backward compatible

chrisdunelm:
This demonstrates upgrading a resource to a resource-set in a backward-compatible way,

the most important files are (sorry that some of the file-names don't quite make sense!):
* ResourceOnly.proto
* ResourceOnlyResourceNames.PlusResourceSets.cs
* SvcClient.PlusResourceSet.cs


chrisdunelm:
Note that this all compiles successfully with the resource + resource_set backward-compatible generated code.

PR #105, opened 2019-07-09
Resource-names update; does not include child_type resource references.

chrisdunelm:


Google.Api.Generator.Tests/WhitespaceFormatterTest.cs
```diff
@@ -237,13 +237,43 @@ public void ForEach(IEnumerable<string> strings)
...
         {
             // Test class that has no members.
         }
+
+        // Test enum.
+        public enum AnEnum
+        {
+            // Enum members with value.
+            A = 1,
+
```
jskeet: I'm not sure the blank lines here are adding to the readability.
Also, would it be worth including tests with XML comments? That's the most likely scenario.

Google.Api.Generator/Generation/ResourceNamesGenerator.cs
```diff
@@ -70,140 +70,402 @@ public ParamProperty(SourceFileContext ctx, string name)
...
-                            Return(New(_ctx.Type(def.One.ResourceNameTyp))(def.One.Template.ParameterNames.Select((_, i) => resourceName.ElementAccess(i)))))
+                    var name = Parameter(_ctx.Type(single.ResourceNameTyp), single.ResourceNameTyp.Name.ToLowerCamelCase());
+                    yield return Method(Public | Static, _ctx.Type(_def.ContainerTyp), "From")(name)
+                        .WithBody(New(_ctx.Type(_def.ContainerTyp))(_ctx.Type(OneOfTypeEnumTyp).Access(single.ResourceNameTyp.Name), name))
                         .WithXmlDoc(
-                            XmlDoc.Summary("Parses the given ", XmlDoc.C(def.DocName), " resource name in string form into a new ", _ctx.Type(def.One.ResourceNameTyp), " instance."),
-                            XmlDoc.Param(name, "The ", XmlDoc.C(def.DocName), " resource name in string form. Must not be ", null, "."),
-                            XmlDoc.Returns("The parsed ", _ctx.Type(def.One.ResourceNameTyp), " if successful.")
-                        );
+                            XmlDoc.Summary("Construct a new instance of ", _ctx.Type(_def.ContainerTyp), " from the provided ", _ctx.Type(single.ResourceNameTyp), "."),
```
jskeet: Construct => Constructs? Or is it "Construct" to match previous code?

Google.Api.Generator/ProtoUtils/ResourceDetails.cs
```diff
@@ -25,148 +26,296 @@ internal class ResourceDetails
...
-                    field.CustomOptions.TryGetMessage<Resource>(ProtoConsts.FieldOption.Resource, out var resource) ? resource : null)
-                    .Where(x => x != null)
-                    .Select(resource => (symbol: MakeFullSymbol(null, msg, resource.Symbol), resource)))
-                .ToList();
-            // Check no duplicates.
-            var duplicateSymbols = resourcesBySymbol1.Select(x => x.symbol).Concat(resourcesBySymbol2.Select(x => x.symbol))
-                .GroupBy(x => x).Where(x => x.Count() > 1).Select(x => x.Key).ToList();
-            if (duplicateSymbols.Any())
+            // Singles.
+            // Compatilibity problems: none
```
jskeet: Compatilibity => Compatibility

Google.Api.Generator/ProtoUtils/ResourceDetails.cs
```diff
@@ -25,148 +26,296 @@ internal class ResourceDetails
...
-                msg.Fields.InFieldNumberOrder().Select(field =>
-                    field.CustomOptions.TryGetMessage<Resource>(ProtoConsts.FieldOption.Resource, out var resource) ? resource : null)
-                    .Where(x => x != null)
-                    .Select(resource => (symbol: MakeFullSymbol(null, msg, resource.Symbol), resource)))
-                .ToList();
-            // Check no duplicates.
-            var duplicateSymbols = resourcesBySymbol1.Select(x => x.symbol).Concat(resourcesBySymbol2.Select(x => x.symbol))
-                .GroupBy(x => x).Where(x => x.Count() > 1).Select(x => x.Key).ToList();
-            if (duplicateSymbols.Any())
+            // Singles.
```
jskeet: Colon instead of period, for consistency with comments below?

Google.Api.Generator.Tests/ProtoTests/Lro/Lro.proto
```diff
@@ -32,13 +32,12 @@ message Request {
...
 
-option (google.api.resource_definition) = {
-  symbol: "Resource",
-  pattern: "items/{item_id}"
-};
-
 message ResourceRequest {
-  string name = 1 [(google.api.resource_reference) = "Resource"];
+  option (google.api.resource) = {
+    type: "lro.example.com/Resource"
```
jskeet: Any chance we could give it a more concrete name, such as "Book"? Things like `class ResourceName : IResourceName` end up suggesting a correlation that isn't really there.
As a reader, I want to see which parts of the generated names are influenced by what's in the proto and which are effectively hard-coded... something like "resource" could be either.
(I'm fine for that to be a change in a separate PR though.)

Google.Api.Generator.Tests/ProtoTests/ResourceNames/ResourceNames.proto
```diff
@@ -11,75 +11,101 @@ import "google/api/field_behavior.proto";
...
 
-  // Test top-level resource definition.
+  // Test simple resource definition.
   rpc SimpleMethod(SimpleResource) returns(Response) {
     option (google.api.method_signature) = "name";
   }
 
-  // Test resource definition inlined in the proto message.
-  rpc SimpleInlineMethod(SimpleInlineResource) returns(Response) {
+  // Test simple resource reference.
```
jskeet: Possibly single rather than simple, to be consistent in terminology everywhere?

Google.Api.Generator.Tests/ProtoTests/ResourceNames/Testing.ResourceNames/ResourceNamesResourceNames.g.cs
```diff
@@ -20,7 +20,7 @@
 
 namespace Testing.ResourceNames
```
jskeet: The fact that this is called ResourceNamesResourceNames.g.cs goes back to my earlier comment about the choice of test names.

Google.Api.Generator.Tests/ProtoTests/ResourceNames/ResourceNames.proto
```diff
@@ -11,75 +11,101 @@ import "google/api/field_behavior.proto";
...
+  // Test multi resource reference.
+  rpc MultiMethodRef(MultiResourceRef) returns(Response) {
+    option (google.api.method_signature) = "multi_resource";
   }
 
-  // Test wilcard resource.
-  rpc WildcardMethod(WildcardResource) returns(Response) {
-    option (google.api.method_signature) = "name,names,required_single,required_repeated";
+  // Test future multi resource (FUTURE_MULTI_PATTERN).
+  rpc FutureMultiMethod(FutureMultiResource) returns(Response) {
```
jskeet: I can't see any methods being generated for this. Possibly I haven't opened a diff? Or is it actually missing?

Google.Api.Generator.Tests/WhitespaceFormatterTest.cs
chrisdunelm: In general in generated code, I think it is better to have blank lines between enum entries, especially given that there will almost always be XmlDoc present.
Unfortunately this whitespace testing doesn't currently support XmlDoc (or any form of comments), so it can't test with xmldoc currently.
I could add this, but would rather do in a future PR.

Google.Api.Generator/Generation/ResourceNamesGenerator.cs
chrisdunelm: Done

Google.Api.Generator/ProtoUtils/ResourceDetails.cs
chrisdunelm: Done

Google.Api.Generator/ProtoUtils/ResourceDetails.cs
chrisdunelm: Done

Google.Api.Generator.Tests/ProtoTests/Lro/Lro.proto
chrisdunelm: OK, yes, I see what you mean.
As you say, this change will be better done in a future PR.

Google.Api.Generator.Tests/ProtoTests/ResourceNames/ResourceNames.proto
chrisdunelm: Yes, I meant "Simple" meaning that this was the simplest case. But "Single" is now better.
Done.

Google.Api.Generator.Tests/ProtoTests/ResourceNames/ResourceNames.proto
chrisdunelm: Well spotted, thank you :)
They were be generated, but were not present in the test file, which doesn't contain the entire generated file.
Done.

Google.Api.Generator.Tests/ProtoTests/ResourceNames/Testing.ResourceNames/ResourceNamesResourceNames.g.cs
chrisdunelm: Yes, will improve naming in a future PR.

chrisdunelm:
Extra commit added. PTAL.

PR #112, opened 2019-07-17
gRPC-service-config based retry support

chrisdunelm:


Google.Api.Generator.Tests/ProtoTests/Basic/Basic.proto
```diff
@@ -4,21 +4,14 @@ package testing.basic;
...
 
-import "google/api/annotations.proto";
 import "google/api/client.proto";
 
 service Basic {
   option (google.api.default_host) = "basic.example.com";
   option (google.api.oauth_scopes) = "scope1,scope2";
 
-  // Test summary text for IdempotentMethod
-  rpc IdempotentMethod(Request) returns(Response) {
```
jskeet: Would it be possible to split this PR into two commits, one of which changes to just use "AMethod" and then one for the gRPC service config part? At the moment I'm struggling to see which changes come from what.

Google.Api.Generator.Tests/ProtoTest.cs
```diff
@@ -227,6 +229,54 @@ void CleanUp(string path)
         [Fact]
         public void UnitTests() => ProtoTestSingle("UnitTests", ignoreCsProj: true, ignoreSnippets: true);
 
+        [Fact]
+        public void GrpcServiceConfig()
+        {
+            using (var grpcServiceConfig = Invoker.TempFile())
```
jskeet: Is there a reason this can't just be a file elsewhere? (Or a resource that you copy to the file system?)

Lots of the other tests use real files, so it feels like this should be able to as well.

Google.Api.Generator/ConfigProtos/service_config.proto
```diff
@@ -0,0 +1,327 @@
+﻿// Copyright 2016 The gRPC Authors
```
jskeet: Presumably eventually this will end up somewhere in github centrally?

Google.Api.Generator.Tests/ProtoTest.cs
chrisdunelm: Done!

Google.Api.Generator/ConfigProtos/service_config.proto
chrisdunelm: Yes.

PR #119, opened 2019-08-08
Camelcaser treats digits as word separators

chrisdunelm:


chrisdunelm:
Yes, it's gradually been getting more convoluted. Maybe later I'll refactor it a bit.

PR #122, opened 2019-08-16
Initial C# dockerization

chrisdunelm:


docker-entrypoint.sh
```diff
@@ -0,0 +1,22 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# THIS SCRIPT IS MEANT ONLY TO BE USED IN THE GAPIC-GENERATOR-CSHARP DOCKER IMAGE
+
+protoc --proto_path=/protos/ --proto_path=/in/ \
+  --plugin=protoc-gen-csharpgapic=/usr/src/gapic-generator-csharp/Google.Api.Generator/bin/Release/netcoreapp2.2/linux-x64/publish/Google.Api.Generator \
+  --csharpgapic_out=/out/ \
+  `find /in/ -name *.proto`
```
jskeet: I tend to use `$(find /in/ -name *.proto)` these days. Can't remember why that's preferred over backticks. You might want to quote `*.proto` too, to avoid glob expansion.

chrisdunelm: done

chrisdunelm:
@lukesneeringer thanks, my mistake - I'll fix in a PR shortly.

PR #127, opened 2019-09-13
Support C# common resource names

chrisdunelm:
These are for when a resource-name class is defined in a common location.


Google.Api.Generator.Tests/ProtoTests/CommonResource/CommonResource.proto
```diff
@@ -0,0 +1,33 @@
...
+
+service CommonResource {
+  option (google.api.default_host) = "commonresource.example.com";
+
+  rpc MethodWithCommonResource(Request) returns(Response) {
+  	option (google.api.method_signature) = "project_name";
+  }
+}
+
+// This is defined as a common reosurce for C#, so will not have a class generated for it.
```
jskeet: Nit: reosurce => resource

(I know you only included this to check that I was paying attention.)

Google.Api.Generator/ConfigProtos/common_resources.proto
```diff
@@ -0,0 +1,14 @@
+﻿syntax = "proto3";
```
jskeet: So will each language have their own proto here, because the common resources need to be specified with different info for each language?
I'm just wondering what that'll look like in github...

Google.Api.Generator/ProtoUtils/ProtoCatalog.cs
```diff
@@ -38,18 +38,22 @@ public ProtoCatalog(string defaultPackage, IEnumerable<FileDescriptor> descs)
...
+        private readonly IImmutableSet<string> _commonUrts;
 
         public MessageDescriptor GetMessageByName(string name) => _msgs[name.Contains('.') ? name : $"{_defaultPackage}.{name}"];
 
         public ResourceDetails.Field GetResourceDetailsByField(FieldDescriptor fieldDesc) => _resourcesByFieldName.GetValueOrDefault(fieldDesc.FullName);
 
         public IEnumerable<ResourceDetails.Definition> GetResourceDefsByFile(FileDescriptor fileDesc) =>
             _resourcesByFileName.GetValueOrDefault(fileDesc.Name, ImmutableList<ResourceDetails.Definition>.Empty);
+
+        public bool IsCommonUrt(string type) => _commonUrts.Contains(type);
```
jskeet: Perhaps IsCommonResourceType? "Urt" isn't terribly obvious unless you already know the context.

Google.Api.Generator.Tests/ProtoTests/CommonResource/CommonResource.proto
chrisdunelm: :)
Done

Google.Api.Generator/ConfigProtos/common_resources.proto
chrisdunelm: C# is the only language that (so far) uses "common resource names", so this isn't (yet) an issue.
If any other language requires it, then potentially this definition could be extended.

Google.Api.Generator/ProtoUtils/ProtoCatalog.cs
chrisdunelm: Done

PR #134, opened 2019-09-18
Downgrade the .NET Core SDK used to 2.2.100

jskeet:
This is to fit in with existing CI build systems which don't support 2.2.400.

When .NET Core SDK 3.0 is released, we'll probably want to do
whatever it takes to move everything over to that, and that should
last us for quite a while...


jskeet:
Please check the Dockerfile part in particular - if you have a simple way of testing it, that would be great.

PR #139, opened 2019-09-25
Support gRPC config; add gapic.sh

chrisdunelm:


docker-entrypoint.sh
```diff
@@ -16,7 +16,26 @@
...
+# Parse options.
+while true; do
+  case "$1" in
+    --grpc-service-config ) GRPC_SERVICE_CONFIG="grpc-service-config=/conf/$2"; shift 2;;
+    --* | +([[:word:][:punct:]]) ) shift ;;
+    * ) break ;;
+  esac
+done
+
+#export CLR_OPENSSL_VERSION_OVERRIDE="1.1"
```
jskeet: Remove comment?

gapic.sh
```diff
@@ -0,0 +1,101 @@
...
+
+CMD="$0"
+
+# Set variables used by this script.
+# All of these are set in options below, and all but $PROTO_PATH are required.
+IMAGE=
+IN=
+OUT=
+PLUGIN_OPTIONS=
+PROTO_PATH=`pwd`
```
jskeet: I've been using `$(pwd)` rather than `pwd`. I believe there's some difference, but I don't remember it.

docker-entrypoint.sh
chrisdunelm: done

gapic.sh
chrisdunelm: This file is mostly copy/paste from go and python versions, so I'd rather leave this as-is so it's as similar to the sources as possible.

jskeet: Sure, that's fine.

PR #141, opened 2019-10-14
Support file-level resource definition annotation

chrisdunelm:


Google.Api.Generator.Tests/ProtoTests/CommonResource/CommonResource.proto
```diff
@@ -15,15 +15,11 @@ service CommonResource {
...
-message Project {
-  option (google.api.resource) = {
+// Also defined as a common resource in CommonResourceConfig.json
+option (google.api.resource_definition) = {
     type: "commonresource.example.com/Project",
 	pattern: "projects/{project}"
-  };
-  string name = 1;
-}
+};
```
jskeet: Is there a reason why pattern is indented further than type here?

Google.Api.Generator/ProtoUtils/ResourceDetails.cs
```diff
@@ -156,13 +156,17 @@ public static IReadOnlyList<Definition> LoadResourceDefinitionsByFileName(IEnume
             // Compatibility problems: Adding a new resource with the same pattern as a single-pattern parent.
             //                         Adding a new resource with the same pattern as one of a multi-pattern parent.
             // TODO: Support new (Sept 2019) `name_descriptor` way of specifying resource-names.
-            var msgs = descs
+            var msgsFromProtoMsgs = descs
```
jskeet: Is the TODO here still valid?

Google.Api.Generator.Tests/ProtoTests/CommonResource/CommonResource.proto
chrisdunelm: Tab/space confusion. Fixed.

Google.Api.Generator/ProtoUtils/ResourceDetails.cs
chrisdunelm: Yes. This is a separate annotation change that's currently still internal-only. I don't know when it's expected to become external.

chrisdunelm:
Thanks. Will merge on green.

PR #148, opened 2019-10-17
Fix bug in enum handling, plus test

chrisdunelm:


Google.Api.Generator.Tests/ProtoTests/MethodSignatures/MethodSignatures.proto
```diff
@@ -89,6 +89,11 @@ message MessageRequest {
   repeated Msg repeated_required = 4 [(google.api.field_behavior) = REQUIRED];
 }
 
+enum TopLevelEnum {
+	DEFAULT = 0;
```
jskeet: That's a lot of whitespace. Outdent a bit?

Google.Api.Generator/Utils/Typ.cs
```diff
@@ -59,7 +59,7 @@ private sealed class Array : Typ
 
         private sealed class FromManual : Typ
         {
-            public FromManual(string ns, string name, bool isEnum = false) => (Namespace, Name, IsEnum) = (ns, name, IsEnum);
+            public FromManual(string ns, string name, bool isEnum = false) => (Namespace, Name, IsEnum) = (ns, name, isEnum);
```
jskeet: Ouch. It really feels like the compiler should have warned you about that.

Google.Api.Generator.Tests/ProtoTests/MethodSignatures/MethodSignatures.proto
chrisdunelm: Tabs :( Done

Google.Api.Generator/Utils/Typ.cs
chrisdunelm: Yes, I'm surprised it didn't warn.

PR #153, opened 2019-10-23
Fully qualify type names where necessary in partial classes

chrisdunelm:


chrisdunelm:
No, the generated code **is** all tested that it compiles.
These extra qualifications are not necessary, but it's easier/simpler to allow them to be generated.
And this PR will introduce the qualification necessary for Scheduler.

PR #154, opened 2019-10-23
Test for fully-qualified type access in proto message partial

chrisdunelm:
Follow-up to #153


chrisdunelm:
:)
It's always correctly handled the situation where the property name (of the resource-name property in the partial class) was the same as the type of that property. Interestingly calling a static method on that type doesn't need to be fully qualified - that is, calling a static method on type X from within a property of name X, also of type X, doesn't need fully qualifying.
This test (and the previous PR) are handling the case where a property is being defined which contains a static method being called on a type that shares a name with a *different* property within the same class.

jskeet:
Right. I *suspect* there's some rule within the C# spec that says "Within a property called Color, the name Color in an expression is treated as the type rather than the property." But I need to find that :)

PR #155, opened 2019-10-24
Support multiple common resource configs

chrisdunelm:


chrisdunelm:
Yes, you're correct that it's an error to specify the same resource name in multi configs.
Multiple files are simply combined - no overrides are happening.

PR #163, opened 2019-11-13
Update to GAX v3, fixing snippets accordingly

jskeet:
(Will assign when I've re-reviewed myself.)


Google.Api.Generator/Generation/SnippetCodeGenerator.cs
```diff
@@ -435,8 +435,8 @@ private IEnumerable<ObjectInitExpr> InitRequest()
                         makeRequest,
                         BlankLine,
                         "// Read streaming responses from server until complete",
-                        ResponseStream.WithInitializer(Response.Access(nameof(ServerStreamingBase<int>.ResponseStream))),
-                        While(Await(ResponseStream.Call(nameof(IAsyncEnumerator<int>.MoveNext))()))(
+                        ResponseStream.WithInitializer(Response.Call(nameof(ServerStreamingBase<int>.GetResponseStream))(Ctx.Type(typeof(CancellationToken)).Access(nameof(CancellationToken.None)))),
```
chrisdunelm: I think there's a slight shortcut you can use: `Ctx.Type<CancellationToken>()`

Google.Api.Generator/Generation/SnippetCodeGenerator.cs
```diff
@@ -484,8 +484,8 @@ private IEnumerable<ObjectInitExpr> InitRequest()
                         BlankLine,
                         "// Create task to do something with responses from server",
                         ResponseHandlerTask.WithInitializer(Ctx.Type<Task>().Call(nameof(Task.Run))(LambdaTyped(null, async: true)(
-                            ResponseStream.WithInitializer(Response.Access(nameof(ServerStreamingBase<int>.ResponseStream))),
-                            While(Await(ResponseStream.Call(nameof(IAsyncEnumerator<int>.MoveNext))()))(
+                            ResponseStream.WithInitializer(Response.Call(nameof(ServerStreamingBase<int>.GetResponseStream))(Ctx.Type(typeof(CancellationToken)).Access(nameof(CancellationToken.None)))),
```
chrisdunelm: (Ditto here)

Google.Api.Generator/Generation/SnippetCodeGenerator.cs
jskeet: Ooh thanks, will do that.

jskeet:
Will merge on green. There's an annoyance where the first commit seems to move a submodule backwards, with a later one moving it forward - but I can't work out how to fix that.

Issue #165, opened 2019-11-18
Keyword parameter names need escaping


jskeet:
I've just regenerated Error Reporting, and we've ended up with:

```csharp
public virtual ReportErrorEventResponse ReportErrorEvent(string projectName, ReportedErrorEvent event, gaxgrpc::CallSettings callSettings = null) =>
```

`event` isn't valid as a parameter name, but `@event` is. Note that this isn't just in the method declaration, but in the use too.
Currently working around with a sed script.

chrisdunelm:
Ah yes, keywords are not handled properly as variable names.
Shouldn't be too much problem to fix.

chrisdunelm:
Fixed by #168

PR #168, opened 2019-11-20
Correctly handle keyword identifier names

chrisdunelm:


chrisdunelm:
This PR is much smaller than it looks - all files named `*.g.cs` are generation results, and only need to be skimmed just to make sure they look sane.

chrisdunelm:
Fixes #165

Google.Api.Generator/RoslynUtils/Keywords.cs
```diff
@@ -0,0 +1,93 @@
...
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Text;
+
+namespace Google.Api.Generator.RoslynUtils
+{
+    internal static class Keywords
+    {
+        // Taken from: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/
+        private static IImmutableSet<string> s_keywords = ImmutableHashSet.Create(
```
jskeet: We could *potentially* include the contextual keywords as well - reason being that while having a parameter named `async` wouldn't hurt the compiler, it might confuse syntax highlighters.

chrisdunelm: Yes I thought of this, but think I'd rather just leave them until we see a real case where it causes a problem.

PR #169, opened 2019-11-20
Properly handle type name collisions when generating in unaliased mode

chrisdunelm:


chrisdunelm:
Yes this will handle the Spanner case. `System` is imported in the spanner snippet source file. So all `System` types, including `Type` will be checked for ambiguity.

PR #177, opened 2019-12-11
Resource-name redesign for vNext

chrisdunelm:
Sorry about this size of this.
Please do check at least a few of generated resource-name classes in the tests, as these are the end result of the change.

And I'm aware some of the resource-name tests are now superfluous because they're now effectively testing the same thing, I'll tidy them up in a later PR.

### DIFF
--- a/Google.Api.Generator.Tests/ProtoTests/ChildResource/Testing.ChildResource/ChildResourceResourceNames.g.cs
+++ b/Google.Api.Generator.Tests/ProtoTests/ChildResource/Testing.ChildResource/ChildResourceResourceNames.g.cs
@@ -22,67 +22,178 @@ namespace Testing.ChildResource
     /// <summary>Resource name for the <c>Project</c> resource.</summary>
     public sealed partial class ProjectName : gax::IResourceName, sys::IEquatable<ProjectName>
     {
-        private static readonly gax::PathTemplate s_template = new gax::PathTemplate("projects/{project}");
-
-        /// <summary>
-        /// Parses the given <c>Project</c> resource name in string form into a new <see cref="ProjectName"/> instance.
-        /// </summary>
-        /// <param name="projectName">The <c>Project</c> resource name in string form. Must not be <c>null</c>.</param>
-        /// <returns>The parsed <see cref="ProjectName"/> if successful.</returns>
-        public static ProjectName Parse(string projectName)
+        /// <summary>The possible contents of <see cref="ProjectName"/>.</summary>
+        public enum ResourceType
         {
-            gax::GaxPreconditions.CheckNotNull(projectName, nameof(projectName));
-            gax::TemplatedResourceName resourceName = s_template.ParseName(projectName);
-            return new ProjectName(resourceName[0]);
+            /// <summary>A resource of an unknown type.</summary>
+            Unknown = 0,
+
+            /// <summary>A resource of type 'project'.</summary>
+            Project = 1
         }
 
+        private static gax::PathTemplate s_project = new gax::PathTemplate("projects/{project}");
+
+        /// <summary>Creates a <see cref="ProjectName"/> containing an unknown resource name.</summary>
+        /// <param name="unknownResourceName">The unknown resource name. Must not be <c>null</c>.</param>
+        /// <returns>
+        /// A new instance of <see cref="ProjectName"/> containing the provided <paramref name="unknownResourceName"/>.
+        /// </returns>
+        public static ProjectName CreateUnknown(gax::UnknownResourceName unknownResourceName) =>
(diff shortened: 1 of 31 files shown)


Google.Api.Generator.Tests/ProtoTests/ChildResource/Testing.ChildResource/ChildResourceResourceNames.g.cs
```diff
@@ -22,67 +22,178 @@ namespace Testing.ChildResource
...
+        /// <returns>
+        /// A new instance of <see cref="ProjectName"/> containing the provided <paramref name="unknownResourceName"/>.
+        /// </returns>
+        public static ProjectName CreateUnknown(gax::UnknownResourceName unknownResourceName) =>
+            new ProjectName(ResourceType.Unknown, gax::GaxPreconditions.CheckNotNull(unknownResourceName, nameof(unknownResourceName)));
+
+        /// <summary>Creates a <see cref="ProjectName"/> with the pattern <c>projects/{project}</c>.</summary>
+        /// <param name="projectId">The <c>Project</c> ID. Must not be <c>null</c> or empty.</param>
+        /// <returns>A new instance of <see cref="ProjectName"/> constructed from the provided ids.</returns>
+        public static ProjectName CreateProject(string projectId) =>
```
jskeet: I wonder whether we need a "Name" suffix on all of these, or at least something slightly different. Calling `CreateProject` sounds like it's going to create a project, for example. I know the class name suggests otherwise. Possibly `ForProject` or `FromProject` would be better?

Google.Api.Generator.Tests/ProtoTests/ChildResource/Testing.ChildResource/ChildResourceResourceNames.g.cs
```diff
@@ -22,67 +22,178 @@ namespace Testing.ChildResource
...
     public sealed partial class ProjectName : gax::IResourceName, sys::IEquatable<ProjectName>
     {
-        private static readonly gax::PathTemplate s_template = new gax::PathTemplate("projects/{project}");
+        /// <summary>The possible contents of <see cref="ProjectName"/>.</summary>
+        public enum ResourceType
+        {
+            /// <summary>A resource of an unknown type.</summary>
+            Unknown = 0,
+
+            /// <summary>A resource of type 'project'.</summary>
```
jskeet: I think we'll need to look carefully at the descriptions here, and possibly the enum name as well. The resource is the same type in all cases, it's just how it's named that differs.

Google.Api.Generator.Tests/ProtoTests/Keywords/Testing.Keywords.Snippets/KeywordsClientSnippets.g.cs
```diff
@@ -30,7 +30,7 @@ public void Method1_RequestObject()
             // Initialize request argument(s)
             Request request = new Request
             {
-                EventAsResourceName = new ResourceName("[ITEM_ID]"),
+                EventAsResourceName = ResourceName.CreateItem("[ITEM_ID]"),
```
jskeet: I think we should still use the constructor for snippets of single-pattern resources. I think it's the more idiomatic form, and will still be *valid* even if the name contains multiple patterns later. I do think it's worth using the factory method for multi-pattern resources, and I don't think it's a problem for the snippets to change at that point.

Google.Api.Generator.Tests/ProtoTests/ChildResource/Testing.ChildResource/ChildResourceResourceNames.g.cs
chrisdunelm: Renamed as `From...`

Google.Api.Generator.Tests/ProtoTests/ChildResource/Testing.ChildResource/ChildResourceResourceNames.g.cs
chrisdunelm: Changed as discussed.

Google.Api.Generator.Tests/ProtoTests/Keywords/Testing.Keywords.Snippets/KeywordsClientSnippets.g.cs
chrisdunelm: Not changed, as discussed.

chrisdunelm:
@jskeet Look in `...//ProtoTests/ResourceNames/Testing.ResourceNames/ResourceNamesResourceNames.g.cs`
A single-pattern resource name is class `SingleResourceName`.
A multi-pattern resource name is class `MultiResourceName`.
Thanks!

PR #181, opened 2019-12-16
Fix bug in nested class handling when outside of a class

### LINKED ISSUE #180: Generator fails for paginated response with nested resource message

chrisdunelm:
Fixes #180


chrisdunelm:
The test that uses a resource-name is also using a string, so that test case is still covered.

PR #187, opened 2019-12-17
Test unit-test-generation when a map field is in a method signature

chrisdunelm:


chrisdunelm:
@jskeet I'll merge this now, as I don't think there's anything controversial.
Unit-tests are only generated for request-methods, not method signatures, so there was nothing to change in the unit-test generator.

PR #194, opened 2020-01-06
Ignore obsolete/deprecated fields when generating snippets and unit-t…

chrisdunelm:
…ests

### DIFF
--- a/Google.Api.Generator/ProtoUtils/ProtoExtensions.cs
+++ b/Google.Api.Generator/ProtoUtils/ProtoExtensions.cs
@@ -25,16 +25,20 @@ namespace Google.Api.Generator.ProtoUtils
 {
     internal static class ProtoExtensions
     {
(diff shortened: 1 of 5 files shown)


Google.Api.Generator/ProtoUtils/ProtoExtensions.cs
```diff
@@ -25,16 +25,20 @@ namespace Google.Api.Generator.ProtoUtils
 {
     internal static class ProtoExtensions
     {
+        private static T ReadOption<T>(IDescriptor desc, string optionName)
+        {
+            // There is no support for reading various standard options; so use reflection.
```
jskeet: Eek, that's pretty nasty. We should look at fixing that in Google.Protobuf...

chrisdunelm: one day...

PR #196, opened 2020-01-06
Generate WithGoogleRequestParam calls instead of WithCallSettingsOverlay

### LINKED ISSUE #195: Use WithGoogleRequestParam instead of WithCallSettingsOverlay
It's simpler and means we have one place to get escaping right.

jskeet:
Additionally, use the null-conditional operator when traversing requests.

Fixes #195.


Google.Api.Generator/Generation/ServiceImplClientClassGenerator.cs
```diff
@@ -130,21 +130,16 @@ IEnumerable<SyntaxNode> PerMethod(MethodDetails method)
...
-                                foreach (var propertyName in header.PropertyNames.Skip(1))
-                                {
-                                    access = access.Access(propertyName);
-                                }
-                                var amp = i == 0 ? "" : "&";
-                                return (FormattableString)$"{amp:raw}{header.EncodedName:raw}={Parens(_ctx.Type(typeof(WebUtility)).Call(nameof(WebUtility.UrlEncode))(access))}";
-                            });
-                            fieldInit = fieldInit.Call(nameof(ApiCall<ProtoMsg, ProtoMsg>.WithCallSettingsOverlay))(
-                                Lambda(request)(_ctx.Type<CallSettings>().Call(nameof(CallSettings.FromHeader))("x-goog-request-params", Dollar(strings.ToArray()))));
+                                access = access.Access(propertyName);
```
chrisdunelm: I think this needs `conditional: true` ?

jskeet: I'm now confused about why the tests were passing. Checking...

jskeet: Very weird. It was passing locally - I wonder whether I then hit "undo". Fixed now.

PR #200, opened 2020-01-07
Don't generate common resource names

jskeet:
It looks like some of the common resource handling underwent bitrot when resource names were redesigned.
It's very possible that this "fix" isn't right, but it's a starting point...


chrisdunelm:
Closing in favour of #202

PR #210, opened 2020-01-21
Update to Google.Protobuf 3.11.2

### LINKED ISSUE #209: Use published Google.Protobuf package
Currently we're on a slightly changed version of Google.Protobuf 3.8.0. We at least need to update that to 3.11.2 - but it would be good if we could use a vanilla release. I'll see what I can do.

jskeet:
This is not pretty, but it does appear to work. I would suggest a VC for anything that's not obvious.

Fixes #209.


jskeet:
Will look into the Travis failure.

Google.Api.Generator/ProtoUtils/ProtoExtensions.cs
```diff
@@ -79,58 +94,66 @@ internal static string RemoveEnumPrefix(string enumName, string valueName)
 
         public static string CSharpName(this EnumValueDescriptor desc) => RemoveEnumPrefix(desc.EnumDescriptor.Name, desc.Name);
 
-        private static IEnumerable<(ulong number, ByteString byteString)> GetRepeated(CustomOptions opts, int field)
+        // *Descriptor.GetOption equivalents, with three differences:
+
```
chrisdunelm: nit: remove blank line?

Google.Api.Generator/ProtoUtils/ProtoExtensions.cs
```diff
@@ -79,58 +94,66 @@ internal static string RemoveEnumPrefix(string enumName, string valueName)
...
+        // - Google.Protobuf code throws an NRE if you try to fetch an option and there are no options in the descriptor.
+        // - Google.Protobuf code throws an NRE if you try to fetch a repeated option and it doesn't exist in the descriptor.
+        // - After obtaining options, we reparse the options with our extension registry - because we can't do we when we create the
```
chrisdunelm: Too many "we"s!!

Google.Api.Generator/ProtoUtils/ProtoExtensions.cs
jskeet: Will do.

Google.Api.Generator/ProtoUtils/ProtoExtensions.cs
jskeet: Whoops, yes :)

PR #211, opened 2020-01-22
Update the generator to not use ServiceEndpoint, which has been removed from GAX

jskeet:


Google.Api.Generator.Tests/ProtoTests/Keywords/Testing.Keywords/KeywordsClient.g.cs
```diff
@@ -116,7 +116,7 @@ public override KeywordsClient Build()
     public abstract partial class KeywordsClient
     {
         /// <summary>The default endpoint for the Keywords service, which is a host of "" and a port of 443.</summary>
-        public static gaxgrpc::ServiceEndpoint DefaultEndpoint { get; } = new gaxgrpc::ServiceEndpoint("", 443);
+        public static string DefaultEndpoint { get; } = ":443";
```
chrisdunelm: This probably isn't the right way to handle a missing default endpoint, but it can be fixed later.

jskeet: Yes, that's probably true. I don't think it's any worse than it was :)

PR #212, opened 2020-01-22
Simplify the abstract client factory methods.

### LINKED ISSUE #207: Review client creation methods
We're currently generating what we used to generate, i.e.

- CreateAsync(gaxgrpc::ServiceEndpoint endpoint = null, XyzSettings settings = null)
- Create(gaxgrpc::ServiceEndpoint endpoint = null, XyzSettings settings = null)
- Create(grpccore::Channel channel, XyzSettings settings = null)
- Create(grpccore::CallInvoker callInvoker, XyzSettings settings = null)

As we're taking a breaking change anyway, and we now have the client builder pattern, we should check that we really want all of those.

jskeet:
We now just expose three factory methods:

- An internal one accepting a CallInvoker and settings; this is as it was before, just internal.
- A parameterless Create method, which uses default credentials, endpoint and settings
- A CreateAsync method accepting an optional cancellation token, which uses default credentials, endpoint and settings

For anything else, we recommend the use of the client builder, which has far more flexibility than any reasonable set of methods could.

Note that the existence of a cancellation token for CreateAsync is new, and IIRC the guts of Google.Api.Auth don't fully support it yet, but it's worth putting in now rather than later.

Fixes #207.

### DIFF
--- a/Google.Api.Generator.Tests/ProtoTests/Basic/Testing.Basic/BasicClient.g.cs
+++ b/Google.Api.Generator.Tests/ProtoTests/Basic/Testing.Basic/BasicClient.g.cs
@@ -120,96 +120,22 @@ namespace Testing.Basic
         internal static gaxgrpc::ChannelPool ChannelPool { get; } = new gaxgrpc::ChannelPool(DefaultScopes);
 
         /// <summary>
-        /// Asynchronously creates a <see cref="BasicClient"/>, applying defaults for all unspecified settings, and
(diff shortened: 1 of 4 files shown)


jskeet:
(This was originally just going to be a prototype PR, but it was pretty simple to do it properly in the end.)

jskeet:
Ping for approval? Had forgotten this hadn't been merged yet. Happy to wait until the resource names PR has been merged.

PR #213, opened 2020-01-29
Update resource-name generation to match the revised specification

chrisdunelm:


Google.Api.Generator.Tests/ProtoTests/ChildResource/Testing.ChildResource/ChildResourceClient.g.cs
```diff
@@ -16,226 +16,2486 @@
...
 
+        /// <summary>
+        /// </summary>
+        /// <param name="ref">
+        /// </param>
+        /// <param name="repeatedRef">
+        /// </param>
+        /// <param name="callSettings">If not null, applies overrides to this RPC call.</param>
+        /// <returns>A Task containing the RPC response.</returns>
+        public virtual stt::Task<Response> TripleParentMethodAsync(ProjectName @ref, scg::IEnumerable<gax::IResourceName> repeatedRef, gaxgrpc::CallSettings callSettings = null) =>
```
jskeet: I can't remember why we have IResourceName overloads here - a TripleName parent is always one of ProjectName, DeptName, OrgName... do we always add an IResourceName overload whenever child_type is specified? Or only when there are multiple parents? (I'd expect this for TripleWildcard, but not TripleParent. I suspect there's a good reason for it - I don't have the doc open right now.)

Google.Api.Generator.Tests/ProtoTests/ResourceNames/ResourceNamesFakes.cs
```diff
@@ -12,44 +12,78 @@
...
     }
 
-    public partial class OriginallySingleResourceRef : ProtoMsgFake<OriginallySingleResourceRef>
-    {
-        public string Resource1 { get; set; }
-        public string Resource2 { get; set; }
-    }
+    public partial class Response : ProtoMsgFake<Response> { }
+
+    //public partial class SingleResource : ProtoMsgFake<SingleResource>
```
jskeet: Can these be removed now? Not sure why they're here.

Google.Api.Generator.Tests/ProtoTests/ChildResource/Testing.ChildResource/ChildResourceClient.g.cs
chrisdunelm: Yes, you're right, these shouldn't be here. The `IResourceName` properties are required in the message partial class, but not as a method overload here. I'll fix this.

Google.Api.Generator.Tests/ProtoTests/ResourceNames/ResourceNamesFakes.cs
chrisdunelm: Oops, yes, an oversight. Will fix.

chrisdunelm:
String pagination is still supported, but for now it doesn't attempt to paginate over a resource-name, it just paginates over the string.
I'd quite like to leave it this way in this PR, and file an issue to work out how to handle this long-term.

For the three affected APIs:
* Pubsub - [`ListTopicSnapshotsResponse`](https://github.com/googleapis/googleapis/blob/b491d07cadaae7cde5608321f913e5ca1459b32d/google/pubsub/v1/pubsub.proto#L298) and [`ListTopicSubscriptionsResponse`](https://github.com/googleapis/googleapis/blob/b491d07cadaae7cde5608321f913e5ca1459b32d/google/pubsub/v1/pubsub.proto#L272). Both will be resource annotated.
* Logging - [`ListLogsResponse`](https://github.com/googleapis/googleapis/blob/b491d07cadaae7cde5608321f913e5ca1459b32d/google/logging/v2/logging.proto#L351). This will be resource annotated.
* Firestore - [`ListCollectionIdsResponse`](https://github.com/googleapis/googleapis/blob/b491d07cadaae7cde5608321f913e5ca1459b32d/google/firestore/v1/firestore.proto#L755). This is *not* a resource.

Google.Api.Generator.Tests/ProtoTests/ChildResource/Testing.ChildResource/ChildResourceClient.g.cs
chrisdunelm: Just thinking that the partial class `IResourceName` properties should probably be `get`-only if there isn't a wildcard specified. When setting the user should always have a strong-typed resource-name; it's only when getting that the type can't be known.

chrisdunelm:
Comments now addressed. PTAL.
Apologies about the unpleasantness of reviewing this PR - as usual, please do look at the generated code, especially for the ResourceName and ChildResource tests - mostly that all the correct signatures are present, and the correct partial-proto-message-class properties are present.

Google.Api.Generator.Tests/ProtoTests/ChildResource/Testing.ChildResource/ChildResourceClient.g.cs
jskeet: Just to clarify this conversation, Chris and I then chatted offline - one common scenario where you might *not* know the type is copying from one object to another, which is why Chris re-added the setter.

Google.Api.Generator.Tests/ProtoTests/Paginated/Testing.Paginated/PaginatedClient.g.cs
```diff
@@ -209,8 +209,8 @@ public abstract class PaginatedClient
         /// </summary>
         /// <param name="request">The request object containing all of the parameters for the API call.</param>
         /// <param name="callSettings">If not null, applies overrides to this RPC call.</param>
-        /// <returns>A pageable sequence of <see cref="ResourceName"/> resources.</returns>
-        public virtual gax::PagedEnumerable<ResourceResponse, ResourceName> ResourcedMethod(ResourceRequest request, gaxgrpc::CallSettings callSettings = null) =>
+        /// <returns>A pageable sequence of <see cref="string"/> resources.</returns>
```
jskeet: It seems odd to see a string referred to as a resource, but it's probably not worth special-casing it.

Google.Api.Generator/ProtoUtils/ResourceDetails.cs
```diff
@@ -174,54 +204,92 @@ string GetShortName(ResourceDescriptor resDesc)
...
-                // It is invalid to ask for a parent that is not already defined.
-                // This may change in the future, to allow auto-generating parent resource-names, but this is not currently allowed.
-                throw new InvalidOperationException(
-                    $"Cannot refer to child-type '{resourceRef.ChildType}' in field {msgDesc.Name}.{fieldDesc.Name} because the child patterns are not already defined in a resource.");
             }
-            throw new InvalidOperationException("type or child_type must be set.");
 
             string ParentPattern(string pattern)
             {
+                if (pattern == "*")
```
jskeet: Ooh, is this something I forgot to do in a previous PR? Sorry.

Google.Api.Generator/RoslynUtils/RoslynBuilder.cs
```diff
@@ -167,14 +167,15 @@ public static PropertyDeclarationSyntax AutoProperty(Modifier modifiers, TypeSyn
...
-                SimpleLambdaExpression(parameter.WithType(null), Block(ToStatements(code).ToArray()));
+            var statements = ToStatements(code).ToArray();
+            var expr = SimpleLambdaExpression(parameter.WithType(null), statements.Length == 1 ? MakeExpr(statements[0]) : Block(statements));
             if (async)
             {
                 expr = expr.WithAsyncKeyword(Token(SyntaxKind.AsyncKeyword));
             }
             return expr;
+            CSharpSyntaxNode MakeExpr(CSharpSyntaxNode c) =>
+                c is ReturnStatementSyntax ret ? ret.Expression : c is ExpressionStatementSyntax exprState ? exprState.Expression : c;
```
jskeet: I'd break after each colon here, so it's a sequence of "condition ? value : " lines... effectively like a switch expression. (Presumably you'd use a switch expression here if you were on C# 8...)

Google.Api.Generator.Tests/ProtoTests/ResourceNames/ResourceNames.proto
```diff
@@ -9,97 +9,69 @@ import "google/api/resource.proto";
...
   }
 }
 
-message SingleResource {
+message SinglePattern {
   option (google.api.resource) = {
-    type: "resourcenames.example.com/SingleResource"
+    type: "resourcenames.example.com/SinglePattern"
 	pattern: "items/{item_id}"
+    name_field: "real_name"
```
jskeet: Is there a reason why pattern is more indented than type and name_field?

Google.Api.Generator.Tests/ProtoTests/Paginated/Testing.Paginated/PaginatedClient.g.cs
chrisdunelm: ack

Google.Api.Generator/ProtoUtils/ResourceDetails.cs
chrisdunelm: I don't think so, it was the simplest way to get multi-parent-including-wildcard-patterns working ;)

Google.Api.Generator/RoslynUtils/RoslynBuilder.cs
chrisdunelm: done

Google.Api.Generator.Tests/ProtoTests/ResourceNames/ResourceNames.proto
chrisdunelm: The whole tabs/spaces thing. Fixed.

chrisdunelm:
OK, everything ready for merging now, I'll leave that to you...

PR #219, opened 2020-02-13
Draft of nested resource handling

jskeet:


Google.Api.Generator/ProtoUtils/ResourceDetails.cs
```diff
@@ -192,6 +193,9 @@ string GetShortName(ResourceDescriptor resDesc)
                 }
                 return typeParts[1];
             }
+
+            IEnumerable<MessageDescriptor> GetMessagesAndSelf(MessageDescriptor message) =>
+                message.NestedTypes.SelectMany(GetMessagesAndSelf).Concat(new[] { message });
```
chrisdunelm: nit: Use `.Append(message)`

jskeet: Ooh, nice, will do.

PR #229, opened 2020-03-04
Update to .NET Core SDK 3.1

jskeet:


jskeet:
I'm confused as to why the tests are hanging - any ideas?

jskeet:
(Trying again with a newer version of the test SDK.)

jskeet:
Okay, this is now passing. The "disable progress when running tests" commit is what unstuck things. Annoying, but we can move on.

Feel free to push back on the generator now generating netcoreapp3.1 tests and snippets - although I think if we want 2.x, it should be 2.1 (LTS) rather than 2.2 ([already out of support](https://dotnet.microsoft.com/platform/support/policy/dotnet-core)).

chrisdunelm:
Thanks for figuring this out :)
Everything at 3.1 is fine.

PR #231, opened 2020-03-11
Allow partial class interception of client builder methods

jskeet:
This will make implementing emulator support simpler, for example.

An alternative possibly design would be to have a partial method
"returning" (via a ref parameter) another builder of the same type.
We could then, then we could call Build or BuildAsync on that
builder (if not null). This would mean only implementing a single
partial method in partial classes - but would prohibit any
customization requiring asynchrony.

The scheme in this commit is still not massively onerous for partial
class implementers, but gives more freedom for future expansion.
Note that all of this can change in the generator later, in that
it's a breaking change for generator consumers, but not for end user
developers, as the public API surface remains the same
(Build/BuildAsync).

### DIFF
--- a/Google.Api.Generator/Generation/ServiceBuilderCodeGenerator.cs
+++ b/Google.Api.Generator/Generation/ServiceBuilderCodeGenerator.cs
@@ -24,6 +24,7 @@ using System.Threading;
 using System.Threading.Tasks;
 using static Google.Api.Generator.RoslynUtils.Modifier;
 using static Google.Api.Generator.RoslynUtils.RoslynBuilder;
+using static Microsoft.CodeAnalysis.CSharp.SyntaxFactory;
 
 namespace Google.Api.Generator.Generation
 {
@@ -49,8 +50,12 @@ namespace Google.Api.Generator.Generation
             using (_ctx.InClass(cls))
             {
                 cls = cls.AddMembers(Settings());
+                cls = cls.AddMembers(InterceptBuild());
+                cls = cls.AddMembers(InterceptBuildAsync());
                 cls = cls.AddMembers(Build());
                 cls = cls.AddMembers(BuildAsync());
+                cls = cls.AddMembers(BuildImpl());
+                cls = cls.AddMembers(BuildAsyncImpl());
                 cls = cls.AddMembers(GetDefaultEndpoint());
                 cls = cls.AddMembers(GetDefaultScopes());
                 cls = cls.AddMembers(GetChannelPool());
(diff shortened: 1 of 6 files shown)


Google.Api.Generator/Generation/ServiceBuilderCodeGenerator.cs
```diff
@@ -63,27 +68,54 @@ private ClassDeclarationSyntax Generate()
...
-            var callInvoker = Local(_ctx.Type<CallInvoker>(), "callInvoker");
+            var client = Local(_ctx.Type(_svc.ClientAbstractTyp), "client");
             return Method(Public | Override, _ctx.Type(_svc.ClientAbstractTyp), "Build")()
                 .WithBody(
-                    This.Call("Validate")(),
-                    callInvoker.WithInitializer(This.Call("CreateCallInvoker")()),
-                    Return(_ctx.Type(_svc.ClientAbstractTyp).Call("Create")(callInvoker, Settings())))
+                    client.WithInitializer(Null),
+                    This.Call("InterceptBuild")(Ref(client)),
+                    Return(BinaryExpression(SyntaxKind.CoalesceExpression, IdentifierName("client"), This.Call("BuildImpl")())))
```
chrisdunelm: Please use `client.NullCoalesce(This.Call("BuildImpl")())`

`NullCoalesce` is an extension method defined in `RoslynExtensions.cs`
(and then the `using static Microsoft.CodeAnalysis.CSharp.SyntaxFactory` can go too)

Google.Api.Generator/Generation/ServiceBuilderCodeGenerator.cs
```diff
@@ -63,27 +68,54 @@ private ClassDeclarationSyntax Generate()
...
 
         private MethodDeclarationSyntax BuildAsync()
         {
             var cancellationToken = Parameter(_ctx.Type<CancellationToken>(), "cancellationToken", @default: Default);
+            var task = Local(_ctx.Type(Typ.Generic(typeof(Task<>), _svc.ClientAbstractTyp)), "task");
+            return Method(Public | Override, _ctx.Type(Typ.Generic(typeof(Task<>), _svc.ClientAbstractTyp)), "BuildAsync")(cancellationToken)
+                .WithBody(
+                    task.WithInitializer(Null),
+                    This.Call("InterceptBuildAsync")(cancellationToken, Ref(task)),
+                    Return(BinaryExpression(SyntaxKind.CoalesceExpression, IdentifierName("task"), This.Call("BuildAsyncImpl")(cancellationToken))))
```
chrisdunelm: ditto

Google.Api.Generator/Generation/ServiceBuilderCodeGenerator.cs
jskeet: Hooray. I was surprised to have to do this, but the discoverability in Roslyn isn't great.

jskeet: Ah - it's not quite this simple because "client" is a statement, not an expression syntax. But it's not too bad...

PR #236, opened 2020-03-25
Update parent pattern handling to cope with singleton resources

jskeet:
This changes the logic for "parent pattern" to be broadly:

- Determine the segments of the pattern by splitting on slash
- Remove either the last two segments (if the last segment is
  a parameter, e.g. "{project}") or just the last segment otherwise
- Join the remaining segments together again with a slash separator

(Note that I'm waiting for confirmation this is all correct from multiple sources before merging.)

### DIFF
--- a/Google.Api.Generator/ProtoUtils/ResourceDetails.cs
+++ b/Google.Api.Generator/ProtoUtils/ResourceDetails.cs
@@ -289,22 +289,22 @@ namespace Google.Api.Generator.ProtoUtils
                     throw new InvalidOperationException("type or child_type must be set.");
                 }
             }
+        }
 
-            string ParentPattern(string pattern)
+        // Note: this could be a local method within LoadResourceReference, but making it a regular
+        // internal method allows it to be unit tested.
+        internal static string ParentPattern(string pattern)
+        {
+            if (pattern == "*")
             {
-                if (pattern == "*")
-                {
-                    // Parent of wildcard is a wildcard.
-                    return "*";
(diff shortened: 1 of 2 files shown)


Google.Api.Generator/ProtoUtils/ResourceDetails.cs
```diff
@@ -289,22 +289,22 @@ string GetShortName(ResourceDescriptor resDesc)
...
-                }
-                return pattern.Substring(0, last2Index + 1);
+                // Parent of wildcard is a wildcard.
+                return "*";
             }
+            var segments = pattern.Split('/');
+            var lastSegment = segments.Last();
+            var segmentsToConsume = lastSegment.StartsWith("{") && lastSegment.EndsWith("}") ? 2 : 1;
+            var parentSegments = segments.Take(segments.Length - segmentsToConsume);
+            return string.Join('/', parentSegments);
```
chrisdunelm: Please can you make this fail with an exception if the parent of `pattern` doesn't exist.

jskeet: Will do, and will add a test for that.

jskeet: Done, PTAL.

jskeet:
Great, thanks. Will merge after internal sign-off on the overall aim.

jskeet:
We now have internal sign-off. Merging.

PR #239, opened 2020-04-15
Support multi-ID path-segments in resource names

chrisdunelm:
No changes required to GAX


Google.Api.Generator/Generation/ResourceNamesGenerator.cs
```diff
@@ -225,7 +251,7 @@ private IEnumerable<MethodDeclarationSyntax> FormatMethods()
                 }
             }
 
-            private IEnumerable<MethodDeclarationSyntax> ParseMethods()
+            private IEnumerable<MemberDeclarationSyntax> ParseMethods()
```
jskeet, line 254: I don't understand why this declaration needs to change - you're not yielding any non-methods, are you?

chrisdunelm: Oh yes, you're right.
For a while it was going to emit an extra field, but I ended up not doing that.
I'll revert this change.

PR #241, opened 2020-04-16
Allow '=**' and '=*' suffix in path-segment variables

chrisdunelm:


Google.Api.Generator.Tests/ProtoTests/ResourceNames/Testing.ResourceNames/ResourceNamesResourceNames.g.cs
```diff
@@ -509,7 +509,7 @@ public enum ResourceNameType
             new WildcardMultiPatternMultipleName(ResourceNameType.ItemA, itemAId: gax::GaxPreconditions.CheckNotNullOrEmpty(itemAId, nameof(itemAId)));
 
         /// <summary>
-        /// Creates a <see cref="WildcardMultiPatternMultipleName"/> with the pattern <c>items_b/{item_b_id}</c>.
+        /// Creates a <see cref="WildcardMultiPatternMultipleName"/> with the pattern <c>items_b/{item_b_id=*}</c>.
```
jskeet, line 512: I suspect it would actually be *nicer* to omit the `=*` and `=**` parts from the documentation, but we definitely don't need to do that now. (It's *almost* an implementation detail.)

Google.Api.Generator/Utils/ResourcePattern.cs
```diff
@@ -52,7 +52,24 @@ public Segment(string segment)
...
+                        {
+                            s = s[0..^3];
+                        }
+                        else if (s.EndsWith("=*"))
+                        {
+                            s = s[0..^2];
+                        }
+                        if (s == "")
+                        {
+                            throw new ArgumentException($"Segment '{segment}' is ill-formed; it is empty.");
```
jskeet, line 66: I suspect we may want to tighten up on terminology when we write the AIP. I tend to think of "segment" as a bit of the pattern between slashes. The bit in braces might be a "placeholder specifier" or similar.

So `~` is valid in segment, but not in a placeholder specifier.

But that can all wait - it's an implementation detail.

Google.Api.Generator.Tests/ProtoTests/ResourceNames/Testing.ResourceNames/ResourceNamesResourceNames.g.cs
chrisdunelm: ack

Google.Api.Generator/Utils/ResourcePattern.cs
chrisdunelm: Yes, agreed that various implemention details need tightening up.
I'm expecting to do this once the spec is finalized, as you suggest :) However, it's good enough for right now.

PR #246, opened 2020-04-27
Update resource-names to conform to the latest specification

chrisdunelm:


jskeet:
Will review this in the morning. Ping me if I forget :)

Google.Api.Generator/Utils/ResourcePattern.cs
```diff
@@ -16,107 +16,154 @@
...
+                    }
+                }
+            }
 
-            public string Expand(IEnumerable<string> parameters) => Separators[0] + string.Join("", parameters.Zip(Separators.Skip(1), (p, s) => p + s));
+            // Separators are between parameter-names; there is always one less separator than parameter-names.
+            public override IReadOnlyList<char> Separators { get; }
+            public override IReadOnlyList<string> ParameterNames { get; }
+            public override IReadOnlyList<string> ParameterNamesWithSuffix { get; }
+            public override string PathTemplateString => $"{{{(ParameterCount == 1 ? ParameterNamesWithSuffix[0] : string.Join('_', ParameterNames))}}}";
```
jskeet, line 122: A comment here would be useful. It took me a while to work out what's going on here. (I understand now.)

Google.Api.Generator/Utils/ResourcePattern.cs
```diff
@@ -16,107 +16,154 @@
...
-                if (!inSep)
-                {
-                    throw new ArgumentException($"Segment '{segment}' is ill-formed; missing final '}}'.");
-                }
-                separators.Add(sb.ToString());
-                if (separators.Any(x => x != "") && parameterNames.Zip(parameterNamesWithSuffix, (a, b) => (a, b)).Any(x => x.a != x.b))
+                if (parameterNames.Count > 1)
                 {
-                    throw new ArgumentException($"Segment '{segment}' is ill-formed; multi-id segment may not use '=*' or '=**' suffix.");
+                    Check(parameterNamesWithSuffix.All(p => !p.EndsWith("=**")), "'=**' suffix not valid in a multi-ID segment.");
```
jskeet, line 103: I don't see how we'd get here anyway - aren't we removing the `=**` in lines 80-83?

Google.Api.Generator/Utils/ResourcePattern.cs
```diff
@@ -16,107 +16,154 @@
...
+            public int ParameterCount => ParameterNames.Count;
+            public bool IsComplex => Separators.Count > 0;
+
+        }
+
+        private class CollectionIdentifierSegment : Segment
+        {
+            public CollectionIdentifierSegment(string segment)
+            {
+                bool valid = segment != "" &&
```
jskeet, line 44: It feels like it might be worth extracting this validation, and the one for the separator characters, so you can reuse them. Otherwise if anything changes, it'll be really easy to only update one...

Google.Api.Generator/Utils/ResourcePattern.cs
```diff
@@ -16,107 +16,154 @@
...
 
-        public string Expand(IEnumerable<string> parameters) => string.Join('/',
+        public string Expand(IReadOnlyList<string> parameters) => string.Join('/',
             Segments.Aggregate((result: ImmutableList<string>.Empty, paramOfs: 0),
                 (acc, seg) => (acc.result.Add(seg.Expand(parameters.Skip(acc.paramOfs))), acc.paramOfs + seg.ParameterCount),
                 acc => acc.result));
 
         public string PathTemplateString => string.Join('/', Segments.Select(x => x.PathTemplateString));
+
+        public string ParentComparisonString => string.Join('/', Segments.Select(seg => seg switch
```
jskeet, line 145: Comment for what this is about, with examples?
Basically this allows `projects/{project}` and `projects/{something_else}` to match each other?

Google.Api.Generator/Utils/ResourcePattern.cs
chrisdunelm: done

Google.Api.Generator/Utils/ResourcePattern.cs
chrisdunelm: This is using the `parameterNamesWithSuffix` list, which _does_ contain the suffixes. This check is only relevant if there is more than one parameter-name, hence it's position in this `if` statement.

Google.Api.Generator/Utils/ResourcePattern.cs
chrisdunelm: None of the validations are quite the same, so I'll leave it as-is for now.
Very happy to extract in a future PR.

Google.Api.Generator/Utils/ResourcePattern.cs
chrisdunelm: Done

PR #247, opened 2020-04-28
Fix resource-names to support the pubsub legacy _deleted-topic_

chrisdunelm:


Google.Api.Generator/Utils/ResourcePattern.cs
```diff
@@ -41,13 +41,21 @@ private class CollectionIdentifierSegment : Segment
         {
             public CollectionIdentifierSegment(string segment)
             {
-                bool valid = segment != "" &&
-                    segment[0] >= 'a' && segment[0] <= 'z' && segment.All(c => (c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z') || (c >= '0' && c <= '9'));
+                Identifier = segment;
+                Validate(tightValidation: false);
+            }
+
+            public void Validate(bool tightValidation)
```
jskeet, line 48: Is this ever called with tightValidation=false? I can only see the call in the ResourcePattern constructor, where it's false.

Google.Api.Generator/Utils/ResourcePattern.cs
```diff
@@ -131,7 +139,19 @@ void Check(bool cond, string msg)
...
+        {
+            Segments = pattern.Split('/').Select(Segment.Create).ToImmutableList();
+            if (Segments.OfType<ResourceIdSegment>().Any())
+            {
+                // Perform tight (standard) validation if there are any resource-id path-segments in this pattern.
+                // Loose validation is required to support pubsub's legacy use of the pattern '_deleted-topic_'
+                foreach (var segment in Segments.OfType<CollectionIdentifierSegment>())
+                {
+                    segment.Validate(tightValidation: true);
+                }
```
jskeet, line 152:
```suggestion
                foreach (var segment in Segments)
                {
                    segment.Validate(tightValidation: segment is CollectionIdentifierSegment);
                }
```

Google.Api.Generator/Utils/ResourcePattern.cs
chrisdunelm: Yes, in the `CollectionIdentifierSegment` constructor it's called with `tightValidation: false`.

Google.Api.Generator/Utils/ResourcePattern.cs
chrisdunelm: This isn't doing quite the same thing - only the `CollectionIdentiferSegment` type has a `Validate(...)` method, so the `OfType<CollectionIdentiferSegment>` in the `foreach` is required to make `segment` the correct type.

Google.Api.Generator/Utils/ResourcePattern.cs
jskeet: Ah, whoops. Hadn't seen that. As you were :)

chrisdunelm:
Agreed, I'll bring this up with Luke.

PR #249, opened 2020-05-07
Use explicit docs instead of inheritdoc in most places

### LINKED ISSUE #248: Generate actual docs instead of using inheritdoc
Docfx is pretty broken with inheritdoc. We use it in ClientBuilder generation, but we could easily just hard-code the right docs. Let's do that.

jskeet:
We still use inheritdoc for ==, != and Equals - those don't seem to
cause docfx issues, for whatever reason.

The documentation isn't always *exactly* what would be inherited -
I've simplified it in a couple of places - but it's close enough.

Fixes #248.


jskeet:
Note to self: merge this at a time autosynth won't be running, so we can then regenerate everything in google-cloud-dotnet in one huge PR.

PR #251, opened 2020-06-10
Initial draft of bazel rules

chrisdunelm:


rules_csharp_gapic/csharp_compiler.bzl
```diff
@@ -0,0 +1,115 @@
...
+    if ctx.attr.runtime:
+        out_run_sh = ctx.actions.declare_file("run.sh")
+        run_sh_contents = """#!/bin/bash
+cd $(dirname $0)
+bin/{configuration}/{framework}/{runtime}/publish/{exe_name}
+""".format(
+            configuration = ctx.attr.configuration,
+            framework = ctx.attr.framework,
+            runtime = ctx.attr.runtime,
+            exe_name = csproj_name[:-7], # remove tailing `.csproj`
```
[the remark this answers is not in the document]
chrisdunelm: Done

rules_csharp_gapic/csharp_compiler_repo.bzl
```diff
@@ -0,0 +1,93 @@
+
```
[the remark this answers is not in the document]
chrisdunelm: Probably not (sadly).
This C# implementation does not cover all use-cases for invoking the compile, which would be a great deal more work.
And we don't want to be responsible for maintaining it.

chrisdunelm:
Thanks @alexander-fenster for the review. And for confirming that it may not be possible to simplify the `WORKSPACE.bazel` content, which is a little disappointing.

rules_csharp_gapic/csharp_gapic.bzl
```diff
@@ -0,0 +1,91 @@
...
+    proto_custom_library(
+        name = name,
+        deps = srcs,
+        plugin = Label("@com_github_grpc_grpc//src/compiler:grpc_csharp_plugin"),
+        output_type = "grpc",
+        output_suffix = ".zip",
+        **kwargs
+    )
+
+def csharp_gapic_library(name, srcs, proto = None, grpc = None, **kwargs):
```
vam-google, line 71: Please let `csharp_gapic_library` rule do just gapic library. The rule, which packs everything together (grpc, proto and gapic), should be called `csharp_gapic_assembly_pkg`, defined in `csharp_gapic_pkg.bzl` file and explicitly used from `BUILD.bazel` file. Please check https://github.com/googleapis/googleapis/blob/master/google/cloud/language/v1/BUILD.bazel#L342  for an example

rules_csharp_gapic/csharp_gapic.bzl
```diff
@@ -0,0 +1,91 @@
...
+        "gapic": attr.label(allow_single_file=True),
+        "_zipper": attr.label(default = Label("@bazel_tools//tools/zip:zipper"), cfg = "host", executable=True),
+    },
+)
+
+def csharp_proto_library(name, srcs, **kwargs):
+    # Build zip file of protoc output
+    proto_custom_library(
+        name = name,
+        deps = srcs,
```
vam-google, line 54: For consistency reasons with the other `<lang>_proto_library` rules, many of which are not defined by us, this argument should be calle `deps` on the surface of the rule itself. Please rename it. See [langauge BUIDL.bazel](https://github.com/googleapis/googleapis/blob/master/google/cloud/language/v1/BUILD.bazel#L320) as an example (all of the `<lang>_proto_library` rules there)

rules_csharp_gapic/csharp_gapic.bzl
```diff
@@ -0,0 +1,91 @@
...
+        output_type = "csharp",
+        output_suffix = ".zip",
+        **kwargs
+    )
+
+def csharp_grpc_library(name, srcs, **kwargs):
```
vam-google, line 60: Please make sure that the surface of this rule matches the following:
```bzl
csharp_grpc_library(
    name = "language_csharp_grpc",
    srcs = [":language_proto"],
    deps = [":language_csharp_proto"],
)
```

This is currently how `<lang>_grpc_library` rules look. For this particular case the `deps` argument will be ignored, but it will make the surface consistent plus it will be used if this rule starts actually compiling the generated output (this is why that argument is present in most of the rules).

rules_csharp_gapic/csharp_gapic.bzl
```diff
@@ -0,0 +1,91 @@
...
+        deps = srcs,
+        plugin = Label("//rules_csharp_gapic:csharp_gapic_generator_binary"),
+        output_type = "gapic",
+        output_suffix = ".zip",
+        **kwargs
+    )
+    if proto and grpc:
+        # Build 'full' output of proto+grpc+gapic output, with all files in the correct locations
+        # for building. This target is named '{name}_full' is created in the 'full/' directory.
+        csharp_gapic_library_full(
```
vam-google, line 84: Please convert csharp_gapic_library_full in `scharp_gapic_assembly_pkg` (see previous comment). Also, please make sure that the surface of the rule matches the other asembly rules.

rules_csharp_gapic/csharp_gapic_repo.bzl
```diff
@@ -0,0 +1,29 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+def _gapic_generator_src_impl(ctx):
+    ctx.download_and_extract(
+        url = "https://github.com/googleapis/gapic-generator-csharp/archive/v1.0.0-beta06.tar.gz",
```
vam-google, line 17: These arguments look like they should be on the sruface of the rule itself, not hardcoded in the implementation (otherwise we would need to update the implementation for each release of gapic-generator-csharp).

rules_csharp_gapic/csharp_compiler_repo.bzl
```diff
@@ -0,0 +1,92 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+def _csharp_compiler_impl(ctx):
+    ctx.download_and_extract(
+        url="https://download.visualstudio.microsoft.com/download/pr/8db2b522-7fa2-4903-97ec-d6d04d297a01/f467006b9098c2de256e40d2e2f36fea/dotnet-sdk-3.1.301-linux-x64.tar.gz",
```
vam-google, line 17: We would need to support at least mac platform as well.  The proper way of doing it would be toolchains, but a simple way would be just checking the os type on the repository rule and downloading a proper binary based on that. Please check PHP implementation for an example of how to do it: https://github.com/googleapis/gapic-generator/blob/master/rules_gapic/php/php_gapic_repositories.bzl#L26

rules_csharp_gapic/BUILD.bazel
```diff
@@ -0,0 +1,25 @@
...
+load(":csharp_compiler.bzl", "csharp_binary")
+
+csharp_binary(
+    name = "csharp_gapic_generator_binary",
+    visibility = ["//visibility:public"],
+    restore_packages = "@gapic_generator_restore//:packages",
+    restore_obj = "@gapic_generator_restore//:obj",
+    src_base = "@gapic_generator_src//:gen_dest/Google.Api.Generator",
+    csproj_name = "Google.Api.Generator.csproj",
+    runtime = "linux-x64",
```
vam-google, line 24: The binary should be built for the host platform automatically (on which the bazel build command is executed). I see that this is an initial implementation of the rules and not everything is handled yet, but the platform should not be an argument of a build rule itself (it undermines many fundamental bazel's principles). Why is this needed?

repositories.bzl
```diff
@@ -0,0 +1,42 @@
...
+    maybe(
+        gapic_generator_src,
+        name = "gapic_generator_src",
+    )
+    maybe(
+        dotnet_restore,
+        name = "gapic_generator_restore",
+        src_base = "@gapic_generator_src//:gen_dest/Google.Api.Generator",
+        csproj_name = "Google.Api.Generator.csproj",
+        runtime = "linux-x64",
```
vam-google, line 41: We should not have platform as an argument to any bazel rules. This basically makes builds on mac fundamentally impossible (even mac-specific part of the rules was implemented this linux-specific portion would still be imported even when executed on mac, resulting to a failure).
For repository rules, the rule itself can determine which platform it is running on like in https://github.com/googleapis/gapic-generator/blob/master/rules_gapic/php/php_gapic_repositories.bzl#L26

rules_csharp_gapic/csharp_gapic.bzl
chrisdunelm: done

rules_csharp_gapic/csharp_gapic.bzl
chrisdunelm: done

PR #253, opened 2020-06-16
Add a readme for bazel use

chrisdunelm:


BAZEL.md
```diff
@@ -0,0 +1,116 @@
...
+    sha256 = "e5265d552e12c1f39c72842fa91d84941726026fa056d914ea6a25cd58d7bbf8",
+    urls = ["https://github.com/protocolbuffers/protobuf/archive/v3.12.3.zip"],
+    strip_prefix = "protobuf-3.12.3",
+)
+load("@com_google_protobuf//:protobuf_deps.bzl", "protobuf_deps")
+protobuf_deps()
+
+http_archive(
+    name = "com_github_grpc_grpc",
+    strip_prefix = "grpc-1.29.1",
```
jskeet, line 22: Can Bazel files include comments? If so, it might be worth explaining that although the major version number here is 1, it's for use with Grpc.Core (etc) 2.x.

BAZEL.md
```diff
@@ -0,0 +1,116 @@
+# BAZEL usage instructions
+
+To build a C# gapic client library from a proto service description,
+the following files are required:
+
+`WORKSPACE.bazel`:
```
jskeet, line 6: Rather than have the content of all these files here in the README, how about having a bazel-example directory that's ready-to-build?

chrisdunelm: Yes, definitely a better idea - I'll abandon this PR and do another...

PR #254, opened 2020-06-16
Bazel example

chrisdunelm:


bazel_example/BUILD.bazel
```diff
@@ -0,0 +1,33 @@
+
```
jskeet, line 1: Is there a reason for the blank line?

chrisdunelm: No, removed!

PR #261, opened 2020-06-23
Add support for grpc-service-config; dep updates

chrisdunelm:
Add support for specifying grpc-service-config to the gapic bazel rule. Update gapic and grpc dependencies. Remove a pointless bazel self-reference to this repo. Update example to include an optional field.


rules_csharp_gapic/BUILD.bazel
```diff
@@ -19,7 +19,7 @@ csharp_binary(
     visibility = ["//visibility:public"],
     restore_packages = "@gapic_generator_restore//:packages",
     restore_obj = "@gapic_generator_restore//:obj",
-    src_base = "@gapic_generator_src//:gen_dest/Google.Api.Generator",
+    src_base = "//:Google.Api.Generator",
```
jskeet, line 22: Is this expected?

chrisdunelm: Yes - previously it was (pointlessly) fetching a second copy of this gapic-generator-csharp repo to use as the source-code for the generator.
This change makes it use *this* copy of the repo.

chrisdunelm:
Thanks :)

PR #262, opened 2020-06-23
Add bazel support for common-resources-config

chrisdunelm:


bazel_example/BUILD.bazel
```diff
@@ -35,6 +39,7 @@ csharp_gapic_library(
         ":example_csharp_grpc",
     ],
     grpc_service_config = "grpc_service_config.json",
+    common_resources_config = "common_resources_config.json",
```
vam-google, line 42: why does it use json instead of common_resources.proto form googleapis?

chrisdunelm: This is an unfortunate naming collision. The "common_resources" referenced here is the JSON representation of [this common_resources.proto](https://github.com/googleapis/gapic-generator-csharp/blob/master/Google.Api.Generator/ConfigProtos/common_resources.proto), which allows multiple C# libraries to share an implementation of a resource.
For example, it allows all C# libraries to use the same "cloudresourcemanager.googleapis.com/Project" implementation from Gax.
Note that we also use the `common_resources.proto` from googleapis exactly as any other language does!

PR #265, opened 2020-07-09
Split Google.Api.Generator into two projects

jskeet:
The new Google.Api.Generator.Utils package contains:

- Formatting utilities
- Roslyn utilities
- System extensions
- The Disposable utility type for "dispose via an action"
- The Typ type which is used widely within Google.Api.Generator

I've moved all the Typ code that's related to protobufs to a new
ProtoTyp class that's still in Google.Api.Generator, to avoid
unnecessary Protobuf dependencies in the Utils project.

While Google.Api.Generator is perhaps not the ideal executable name
now, it's not *too* bad, and it's probably not worth the widespread
inconvenience of changing it.


jskeet:
cc @amanda-tarafa

jskeet:
Woot! Thanks... who knows, by tomorrow we may have the skeleton of the Discovery-based generator...

PR #270, opened 2020-09-01
Fix test and snippet generation for constant patterns

jskeet:
This involves unfortunate duplication, but I suspect that can't be
avoided without a bit more disruption. (Ideally the
SnippetCodeGenerator, UnitTestCodeGenerator and
could use ResourceNamesGenerator as the one true source of
information about what would be generated, but that would be a big
change.)

No tests yet - there's no *simple* place to add a test for this.
Suggestions welcome, including "it's not worth it for this case".

### DIFF
--- a/Google.Api.Generator/Generation/SnippetCodeGenerator.cs
+++ b/Google.Api.Generator/Generation/SnippetCodeGenerator.cs
@@ -177,7 +177,7 @@ namespace Google.Api.Generator.Generation
                     {
                         @default = pattern is null ?
                             (object)New(Ctx.Type<UnparsedResourceName>())("a/wildcard/resource") :
-                            Ctx.Type(resourceTyp).Call($"From{string.Join("", pattern.Template.ParameterNames.Select(x => x.RemoveSuffix("_id").ToUpperCamelCase()))}")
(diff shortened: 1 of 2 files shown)


jskeet:
cc @amanda-tarafa for context.

chrisdunelm:
Yes, really this needs a refactoring to allow the UnitTests and Snippets to use the helpers in `ResourceNamesGenerator`, as there's code there which has already sorted out the correct naming.
But I agree that this PR is suitable until that refactor is done.

To test, I think the simplest way would be to change the `pattern: "items/{item_id}"` in the `WildcardMultiPattern` message in `ResourceNames.proto`; such that the first pattern is a constant pattern (ie without any parameters).
This would test the required case, without requiring any large extra tests.
And the `SinglePattern` test still tests the case of the first pattern (only pattern in this case) being a standard pattern with a parameter.

Alternatively, add an extra RPC and message to this `ResourceNames.proto` test, specifically to test this case. But this will produce a lot more test output (which isn't itself a problem exactly, but this test is already fairly verbose!)

chrisdunelm (approved):
LGTM
Test to follow in a future PR.

PR #271, opened 2020-09-02
Enable snippets and unittests for resource-name tests

chrisdunelm:
This does not yet test the new 'parameterless pattern' code; that change will be in a follow-up PR


jskeet (approved):
Thanks!

PR #273, opened 2020-09-09
More REST generator changes

jskeet:
Review one commit at a time. This is just fixing up the base request, really - although I expect a lot of it may be useful elsewhere later.

### DIFF
--- a/Google.Api.Generator.Rest/Models/EnumMemberModel.cs
+++ b/Google.Api.Generator.Rest/Models/EnumMemberModel.cs
@@ -1,0 +1,48 @@
+﻿// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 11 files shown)


chrisdunelm:
One slightly nitty point... There is a change to the white-space formatter but no test for it; it a test attribute with args could be added that would be good :)

jskeet:
Yup, good point - will add that as an extra commit.

jskeet:
@chrisdunelm: See the new last commit. It's a bit odd, to be honest...

chrisdunelm (approved):
Yes, I see what you mean.
There may be a more elegant solution, but as you say, this works for now :)
LGTM

jskeet:
Cool - will wait for Amanda to look tomorrow before merging, mostly for the REST aspect.

amanda-tarafa (approved):
LGTM

PR #274, opened 2020-09-10
More REST generator progress

jskeet:


Google.Api.Generator.Rest/Models/MethodModel.cs
```diff
@@ -15,25 +15,84 @@
...
+        public Typ RequestTyp { get; }
         public string Name { get; }
 
-        public MethodModel(PackageModel package, string name, RestMethod restMethod)
+        public MethodModel(PackageModel package, ResourceModel resource, string name, RestMethod restMethod)
         {
             Package = package;
+            Resource = resource;
             Name = name.ToUpperCamelCase();
+            ParentTyp = resource?.Typ ?? package.ServiceTyp;
```
amanda-tarafa, line 46: Oh, I din't know that the methods could be defined outside of the resources, directly on the service, don't remember having seen that.

Google.Api.Generator.Rest/Models/MethodModel.cs
```diff
@@ -15,25 +15,84 @@
...
+        public IReadOnlyList<ParameterModel> Parameters { get; }
+        public Typ ParentTyp { get; }
+        public Typ RequestTyp { get; }
         public string Name { get; }
 
-        public MethodModel(PackageModel package, string name, RestMethod restMethod)
+        public MethodModel(PackageModel package, ResourceModel resource, string name, RestMethod restMethod)
         {
             Package = package;
+            Resource = resource;
```
amanda-tarafa, line 44: I think you are not using this anywhere else, I'd say, remove it, if it's not relevant for MethodModel then it shouldn't be relevant outside, or if it becomes relevant then maybe MethodModel should encapsulate that.

Google.Api.Generator.Rest/Models/MethodModel.cs
```diff
@@ -15,25 +15,84 @@
...
+        }
+
+        private static IReadOnlyList<ParameterModel> CreateParameterList(PackageModel package, RestMethod restMethod)
+        {
+            var parameterOrder = restMethod.ParameterOrder ?? new List<string>();
+            // TODO: Skip the Alt parameter
+            // TODO: Handle the body parameter if restMethod.Request is non-null
+            return (restMethod.Parameters ?? new Dictionary<string, JsonSchema>())
+                .Select(pair => new ParameterModel(pair.Key, pair.Value))
+                .OrderBy(p => !p.IsRequired)
```
amanda-tarafa, line 59: I think you can directly skip here the ones that are not required. Or are you planning to be able to generate overloads later?

amanda-tarafa (approved):
LGTM, just a couple of comments.

Google.Api.Generator.Rest/Models/MethodModel.cs
jskeet: I don't have any concrete examples of it at the moment... we might be able to remove it later.

Google.Api.Generator.Rest/Models/MethodModel.cs
jskeet: I'm generally being fairly lax about what to define as properties. It makes things a lot quicker for development. I suggest that we don't worry about it for the moment, but that we go back over all the models later to see what we actually need to expose. It's definitely inconsistent right now.

Google.Api.Generator.Rest/Models/MethodModel.cs
jskeet: Not overloads, but they become properties on the request class - we'll definitely want them.

jskeet:
(Merging as-is.)

PR #275, opened 2020-09-11
Add the remainder of the Request class generated for methods

jskeet:
The next step will be to generate the Data classes properly so we can access those for both the "body" parameter and the response type for request classes.


amanda-tarafa (approved):
LGTM

amanda-tarafa:
Travis is failing though.

jskeet:
That's an odd failure. Restarted... we'll see what happens.

jskeet:
Looks okay now. Merging.

PR #278, opened 2020-09-16
More REST data model population

jskeet:
With these in place, we're getting close to complete. There are still features missing, primarily media upload/download handling... but it's mostly tidy-up of odd little naming issues.


amanda-tarafa (approved):
Yep, all looks good!

PR #279, opened 2020-09-16
Rewrite C# bazel rules

chrisdunelm:
Required to run the latest generator version due to recent use of project-reference


vam-google (approved):
LGTM

Also it looks much shorter and cleaner than the original implementation.

PR #280, opened 2020-09-17
Finish initial generator implementation

jskeet:
There's one annoyance with the feature list formatting, but apart from that, Translate, WebFonts and Storage all generate in a whitespace-only-diff (and copyright) fashion. (For the C# code. We don't generate the project files yet.)

There may well be more differences to take account of when we generate more APIs...

Note that I'm assuming that https://github.com/googleapis/google-api-dotnet-client/pull/1651 will be merged - the final Storage sample update is based on that.


jskeet:
Will investigate the test failure. I *suspect* it's around XML doc generation for lists

PR #282, opened 2020-10-05
More REST generator fixes

jskeet:
This is a bunch of miscellaneous fixes. Once this is in, there's still just a *little* bit of work to do on how data model classes are referenced, and then the C# part will be done. After that, we'll look at project files (which shouldn't be too bad).

### DIFF
--- a/test-all-rest.sh
+++ b/test-all-rest.sh
@@ -1,0 +1,49 @@
+#!/bin/sh
+
+set -e
+
+# Generates all REST APIs present in a google-api-dotnet-client repo
+# using Google.Api.Generator.Rest, and compares the results with the
+# Python-generated code.
+
+if [[ -z "$1" ]]
+then
+  echo "Required argument: path to root directory of google-api-dotnet-client repo"
+  exit 1
+fi
+
+declare -r CLIENT_REPO=$1
+
+rm -rf tmp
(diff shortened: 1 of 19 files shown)


test-all-rest.sh
```diff
@@ -0,0 +1,49 @@
...
+  dotnet run --no-build -p Google.Api.Generator.Rest -- $json tmp/csharp
+done
+
+# Copy the Python-generated code
+for package_dir in tmp/csharp/*
+do
+  package=$(basename $package_dir)
+  if [[ ! -d $CLIENT_REPO/Src/Generated/${package} ]]
+  then
+    echo "Python directory for ${package} does not exist; deleting C# code"
```
amanda-tarafa, line 35: But also in the end, we'd need to check that we have generated everything that python had generated, right? Mostly to protect against the generator not propagating an exception somewhere I think.

amanda-tarafa (approved):
LGTM
Yep, these all look fine to me.

test-all-rest.sh
jskeet: Yup - this only reduces unnecessary diffs due to things we choose not to generate in Python. Note that by now we've run the C# generator already, so if that throws, we'll have found it out already.

PR #283, opened 2020-10-07
Generate csproj files more fully

jskeet:
The generated files are identical to the Python-generated ones.

### DIFF
--- a/Google.Api.Generator.Rest/CodeGenerator.cs
+++ b/Google.Api.Generator.Rest/CodeGenerator.cs
@@ -143,7 +143,12 @@ namespace Google.Api.Generator.Rest
         private static ResultFile GenerateProjectFile(PackageModel package)
         {
             var doc = package.GenerateProjectFile();
-            return new ResultFile($"{package.PackageName}/{package.PackageName}.csproj", doc.ToString());
+            // Take care of fine-grained line-break placement, and add a final line break for compatibility
+            // with existing generator (which really doesn't need to be changed for this).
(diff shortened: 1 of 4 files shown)


Google.Api.Generator.Rest/CodeGenerator.cs
```diff
@@ -143,7 +143,10 @@ void InsertLine(string contentToFind, int offset, string extraLine)
         private static ResultFile GenerateProjectFile(PackageModel package)
         {
             var doc = package.GenerateProjectFile();
-            return new ResultFile($"{package.PackageName}/{package.PackageName}.csproj", doc.ToString());
+            // Take care of fine-grained line-break placement, and add a final line break for compatibility
+            // with existing generator (which really doesn't need to be changed for this).
+            string text = doc.ToString().Replace("  <!--linebreak-->", "") + Environment.NewLine;
```
amanda-tarafa, line 148: You are replacing `linebreak ` by and empty string? Should it be replaced by a new line instead?

amanda-tarafa (approved):
LGTM

Google.Api.Generator.Rest/CodeGenerator.cs
jskeet: No, because the XComment is already formatted onto a line on its own. Sounds hokey, but it works. I'll clarify the comment though :)

amanda-tarafa: Ah yes, makes sense.

PR #284, opened 2020-10-09
Final changes to REST generator for parity

jskeet:
(No urgency to review this)

With these changes in, we generate the same code that the Python generator does.

Next step will be to refactor a bit, ideally to move everything around naming to one place, and reducing the number of public properties.


jskeet:
Urgh - have found one remaining problem, in the filename for the BigQuery API. Will handle that later...

amanda-tarafa (approved):
LGTM

PR #291, opened 2020-11-25
First stages of refactoring

jskeet:
This is primarily about making properties private where possible (or removing them if they're only used once and it can be inlined) and giving more documentation. Definitely not the end, but progress.


amanda-tarafa (approved):
LGTM

PR #293, opened 2020-12-01
fix: rerun dotnet compiler if the first invocation fails

alexander-fenster:
A lot of continuous integration builds (one or two per day) fail in Kokoro for no reason with this error:
```
failed: error executing command bazel-out/host/bin/external/com_google_protobuf/protoc --experimental_allow_proto3_optional '--plugin=protoc-gen-gapic=bazel-out/host/bin/external/gapic_generator_csharp/rules_csharp_gapic/run.sh' ... (remaining 17 argument(s) skipped)

Use --sandbox_debug to see verbose messages from the sandbox
Failed to create CoreCLR, HRESULT: 0x80004005
--gapic_out: protoc-gen-gapic: Plugin failed with status code 137.
```

The latest example of this failure can be seen in Sponge with ID 4a0cfd5f-e352-4b26-94a6-bf56052d59aa. They always pass if I restart the build.

Until the root cause is found (if it's possible to find the root cause at all :) ), let's try this simple hack to run the `dotnet` binary for the second time if the first invocation fails. What do you folks think?

### DIFF
--- a/rules_csharp_gapic/csharp_compiler.bzl
+++ b/rules_csharp_gapic/csharp_compiler.bzl
@@ -25,9 +25,11 @@ DOTNET_CLI_HOME="$(pwd)/local_tmp" \
 DOTNET_SKIP_FIRST_TIME_EXPERIENCE=1 \
 DOTNET_CLI_TELEMETRY_OPTOUT=1 \
 DOTNET_NOLOGO=1 \
-{csharp_compiler}/dotnet build src/{csproj_relative} \
+cmd="{csharp_compiler}/dotnet build src/{csproj_relative} \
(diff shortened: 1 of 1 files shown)


jskeet:
I don't know Bazel so can't comment on the details... I don't object to trying this, but I think we should keep an issue open somewhere to work out the root cause. This is pretty odd.

alexander-fenster:
@chrisdunelm @vam-google ping on this :)

vam-google (approved):
I think it is worth a try, LGTM.

chrisdunelm (approved):
LGTM

PR #294, opened 2020-12-16
Correctly detect stdin with early eof

### LINKED ISSUE #287: InvocationTest.BadCmdLineInvocation fails in GitHub Actions
(The same test passes on Travis.)

It ends up with an exit code of 0 instead of 1. I suspect that's because it detects empty input and responds "to protoc" with the error, rather than hanging *waiting* for input - I don't know why, but I don't believe it's a problem with the generator itself, so I'll skip it for now.

Assigning to Chris to think about further in terms of what the test should do.

chrisdunelm:
Fixes #287


jskeet (approved):
Thanks!

PR #297, opened 2021-01-07
Fix admin apis

jskeet:


test-all-rest.sh
```diff
@@ -40,22 +40,11 @@ done
...
-  else
-    echo "Copying Python code for ${package}"
-    mkdir -p tmp/python/$package
-    cp $CLIENT_REPO/Src/Generated/${package}/*.cs tmp/python/$package
-    cp $CLIENT_REPO/Src/Generated/${package}/*.config tmp/python/$package
-    cp $CLIENT_REPO/Src/Generated/${package}/*.csproj tmp/python/$package
-  fi
-done
+# We don't generate IdentityToolkit
+rm -rf tmp/csharp/Google.Apis.IdentityToolkit.v3
```
amanda-tarafa, line 44: There are some other ones that we don't _currently_ generate because the generated code doesn't build or has some other issues stemming from discovery docs issues. But I think it wouldn't hurt to include them in the diff, presumably the bad code should also be the same, but if it's not, it's also not a problem in my view, given that the issues stem from discovery. Basically, I think it's OK if generators behave differently when discovery is broken.
Just flagging in case you see something like that, now or in future diff executions.

amanda-tarafa (approved):
LGTM

test-all-rest.sh
jskeet: I think the script deletes the discovery doc in that case, so it doesn't show up as a problem in the diff. Note that we're only diffing against the Python code that has already been generated - I don't run the generator separately. The original intention was that the diff would be part of CI, but we probably don't need that now. (I might see how hard it would be to add it...)

amanda-tarafa: Ah yes, if you are diffing against already python generated code, then yes, that wouldn't show, I hadn't realized. And yes, the top level script deletes the discoveries before generation, but not this one, I wasn't sure how you were running this one.

PR #298, opened 2021-01-08
Each snippet is now generated in and individual file (also).

amanda-tarafa:
Existing, single file generation continues until standalone generation is cleanned and standalone snippets are the ones used for docs.
The second commit is big, but it consists mainly of splitting the existing expected output files for snippet into single files, the generate method code hasn't changed. It's probably worth it to review just a sample of those files.
More PRs are comming after this one for adding region tags, some comments, etc. But better to review this one on its own, which is the one doing the per file splitting.
I'll share internally the design doc shortly.

### DIFF
--- a/Google.Api.Generator/Generation/SnippetCodeGenerator.cs
+++ b/Google.Api.Generator/Generation/SnippetCodeGenerator.cs
@@ -15,8 +15,8 @@
 using Google.Api.Gax;
 using Google.Api.Gax.Grpc;
 using Google.Api.Generator.ProtoUtils;
-using Google.Api.Generator.Utils.Roslyn;
 using Google.Api.Generator.Utils;
+using Google.Api.Generator.Utils.Roslyn;
 using Google.LongRunning;
 using Google.Protobuf;
 using Google.Protobuf.Reflection;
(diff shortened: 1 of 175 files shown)


jskeet:
Will wait until I've had a look at the design doc to review - and wouldn't have time before Monday anyway. Added it to my to-do list though.

amanda-tarafa:
Yep, that's totally fine.

Google.Api.Generator/Generation/SnippetCodeGenerator.cs
```diff
@@ -35,102 +35,137 @@ namespace Google.Api.Generator.Generation
...
+
+            public MethodDeclarationSyntax GenerateMethod(SourceFileContext ctx) => MethodGenerator(ctx);
+
+            public NamespaceDeclarationSyntax StandaloneSnippet(SourceFileContext ctx)
+            {
+                var ns = Namespace(Svc.SnippetsNamespace);
+                using (ctx.InNamespace(ns))
+                {
+                    var cls = Class(Public | Sealed | Partial, Svc.StandaloneSnippetsTyp)
+                        .WithXmlDoc(XmlDoc.Summary("Generated snippet."));
```
jskeet, line 150: As noted in the design doc, I suspect we don't want a summary per partial class.

Google.Api.Generator/Generation/SnippetCodeGenerator.cs
```diff
@@ -35,102 +35,137 @@ namespace Google.Api.Generator.Generation
...
+
+        private SnippetDef Snippet { get; }
+        public string SnippetName => Snippet.SnippetName;
+
+        public CompilationUnitSyntax Generate(SourceFileContext ctx) =>
+            ctx.CreateCompilationUnit(Snippet.StandaloneSnippet(ctx));
+
+        private class SnippetDef
+        {
+            public string SnippetName { get; }
```
jskeet, line 135: What is the "name" of a snippet? I can think of a bunch of things this could refer to - a comment with more details and an example would be really helpful.

Google.Api.Generator/Generation/SnippetCodeGenerator.cs
```diff
@@ -540,26 +595,31 @@ public class Signature
                             f.Descs.Last().CSharpFieldName())
                         .WithInitializer(_def.DefaultValue(f.Descs.Last(), null, topLevel: true)));
                 private IEnumerable<LocalDeclarationStatementSyntax> InitRequestArgsNormal => InitRequestArgs(resourceNameAsString: true);
-                private IEnumerable<LocalDeclarationStatementSyntax> InitRequestArgsResourceNames => InitRequestArgs(resourceNameAsString: false);
```
jskeet, line 543: Were these unused?

jskeet (approved):
Some nits/questions, but nothing major. I looked through the second commit to some extent, but certainly not exhaustively - it's not clear to me yet which snippets are generated "only in the existing format", if any, for example.

Google.Api.Generator/Generation/SnippetCodeGenerator.cs
```diff
@@ -35,102 +35,137 @@ namespace Google.Api.Generator.Generation
...
+            private Func<SourceFileContext, MethodDeclarationSyntax> MethodGenerator { get; }
+
+            public SnippetDef(string snippetName, ServiceDetails svc, Func<SourceFileContext, MethodDeclarationSyntax> methodGenerator) =>
+                (SnippetName, Svc, MethodGenerator) = (snippetName, svc, methodGenerator);
+
+            public MethodDeclarationSyntax GenerateMethod(SourceFileContext ctx) => MethodGenerator(ctx);
+
+            public NamespaceDeclarationSyntax StandaloneSnippet(SourceFileContext ctx)
+            {
+                var ns = Namespace(Svc.SnippetsNamespace);
```
amanda-tarafa, line 146: Note to self: As proposed on the design doc, consider changing this namespace to something like Google.Cloud.StandaloneSnippets.AutomMl.V1 (for each API). An update the design doc afterwards.

amanda-tarafa (commented):
I've addressed review comments. PTAL.

jskeet:
Will try to have a look later this afternoon.

amanda-tarafa:
No worries, I'm still not done with the internal doc, so you might as well wait for that to be done.

amanda-tarafa:
Chatter offline. Fine to merge as is. Merging...

PR #300, opened 2021-02-03
Add missing license headers to bazel_example files

chrisdunelm:


chrisdunelm:
I've used "2020" for the year, as these headers should have been in pace last year. Please let me know if you think they should be 2021.

amanda-tarafa (approved):
LGTM

PR #301, opened 2021-02-08
Allow REST libraries to override the major version number

jskeet:
(Review each commit separately)


jskeet:
Not sure why it's failing to build... on it.

jskeet:
Ah, it's the test code - for some reason VS hadn't mentioned that before. Fixing.

jskeet:
(Fixed in the third commit - the second commit would still be broken, but we'll never need to go back to that...)

amanda-tarafa (approved):
LGTM

PR #302, opened 2021-02-11
Proposal: feat: Unaliased source context with own namespace

amanda-tarafa:


amanda-tarafa:
@jskeet would this be OK for being able to include the client namespace in snippets, so that it shows to users? The caveat is what you said before, if a user opens our files they will see a greyed out using statement.
Feasable alternatives are:

- A constant namespace for all standalone snippets.
- No namespace for all standalone snippets.

jskeet:
I don't mind this as a change if it *only* affects snippets (which I believe is the intention).

amanda-tarafa:
Closing as changes in #303 simplify this a lot. Will create a new PR shortly.

PR #304, opened 2021-02-22
Use newer version of protobuf to simplify matters

jskeet:


jskeet:
I'm pretty confident about this change, but when we've merged it I'll regenerate all the APIs locally as an extra test. (That's feasible to do without merging, but more fiddly.)

chrisdunelm (approved):
LGTM

PR #308, opened 2021-03-15
feat: Standalone snippets always import own namespace.

amanda-tarafa:


amanda-tarafa:
First commit is the production code change.
Second commit is the change in the expected output for tests.

Google.Api.Generator/CodeGenerator.cs
```diff
@@ -138,7 +138,7 @@ internal static class CodeGenerator
                     // Generate snippets for the service
                     // TODO: Consider removing this once we have integrated the standalone snippets
                     // with docs generation.
-                    var snippetCtx = SourceFileContext.CreateUnaliased(clock, s_wellknownNamespaceAliases, s_avoidAliasingNamespaceRegex);
+                    var snippetCtx = SourceFileContext.CreateUnaliased(clock, s_wellknownNamespaceAliases, s_avoidAliasingNamespaceRegex, true);
```
jskeet, line 141: It might be worth using named arguments both here and for the other call, just for clarity.

jskeet (approved):
One style-only suggested change, but otherwise looks good.

Google.Api.Generator/CodeGenerator.cs
amanda-tarafa: Yes, done.

amanda-tarafa (commented):
Comment addressed. I'll squash the second commit and wait for autosynth to be done before merging.

PR #312, opened 2021-03-17
Fail generation if duplicate resource names are generated

jskeet:
Tested with AI Platform before and after proto patching.

Before the proto patch, it fails with an error message of:

> System.InvalidOperationException: The following resource name classes were created multiple times: ModelName, DatasetName

Also run the generator over all APIs in google-cloud-dotnet: no differences reported.


amanda-tarafa (approved):
LGTM

viacheslav-rostovtsev (approved):

PR #313, opened 2021-03-18
(REST) Make DownloadRange use the MediaDownloader

jskeet:
This will allow users to subscribe to progress (and make other
configuration changes) when performing ranged downloads.

Tested by:

- Regenerating all APIs in google-api-dotnet-client (succeeds)
- Rebuilding them all (succeeds)
- Checking the changes (all MediaDownloader-related)
- Building a local Storage package
- Using the ProgressChanged event handler as expected:

```csharp
var obj = client.Objects.Get(bucketName, objectName);
obj.MediaDownloader.ProgressChanged += MediaDownloader_ProgressChanged;
obj.MediaDownloader.ChunkSize = 5000;
obj.DownloadRange(new MemoryStream(), new System.Net.Http.Headers.RangeHeaderValue(100, 50_000_000));
```

This will address https://github.com/googleapis/google-api-dotnet-client/issues/1784

Strictly speaking it's a breaking change in that if someone were already modifying the MediaDownloader then using the DownloadRange method, they'll see the difference - but it's much more consistent than before, and I'd be very surprised to see it cause an issue.

### DIFF
--- a/Google.Api.Generator.Rest/Models/MethodModel.cs
+++ b/Google.Api.Generator.Rest/Models/MethodModel.cs
@@ -219,35 +219,53 @@ namespace Google.Api.Generator.Rest.Models
             var cancellationTokenWithDefault = Parameter(ctx.Type<CancellationToken>(), "cancellationToken", DefaultExpression(ctx.Type<CancellationToken>()));
             var range = Parameter(ctx.Type<RangeHeaderValue>(), "range");
 
-             var syncDownloadWithStatus = Method(Modifier.Public | Modifier.Virtual, ctx.Type<IDownloadProgress>(), "DownloadWithStatus")(stream)
-                .WithBlockBody(Return(mediaDownloader.Call("Download")(ThisQualifiedCall("GenerateRequestUri")(), stream)))
+            var downloadRemarks = XmlDoc.Remarks("This method uses the ", mediaDownloader, " property to perform the download. " +
+                "Progress event handlers and other configuration may be performed using that property prior to calling this method.");
+
+            var mediaDownloaderLocal = Local(ctx.Type(Typ.Var), "mediaDownloader")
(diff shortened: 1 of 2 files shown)


Google.Api.Generator.Rest/Models/MethodModel.cs
```diff
@@ -219,43 +219,61 @@ private static ClassDeclarationSyntax AddMediaDownloadMethods(PropertyDeclaratio
             var cancellationTokenWithDefault = Parameter(ctx.Type<CancellationToken>(), "cancellationToken", DefaultExpression(ctx.Type<CancellationToken>()));
             var range = Parameter(ctx.Type<RangeHeaderValue>(), "range");
 
-             var syncDownloadWithStatus = Method(Modifier.Public | Modifier.Virtual, ctx.Type<IDownloadProgress>(), "DownloadWithStatus")(stream)
-                .WithBlockBody(Return(mediaDownloader.Call("Download")(ThisQualifiedCall("GenerateRequestUri")(), stream)))
+            var downloadRemarks = XmlDoc.Remarks("This method uses the ", mediaDownloader, " property to perform the download. " +
```
amanda-tarafa, line 222:
```suggestion
            var downloadRemarks = XmlDoc.Remarks("This method uses the ", MediaDownloader, " property to perform the download. " +
```

amanda-tarafa (approved):
LGTM

Google.Api.Generator.Rest/Models/MethodModel.cs
jskeet: `mediaDownloader` here is a variable representing the property declaration - the `XmlDoc.Remarks` method automatically turns that into a `<see>` link.

amanda-tarafa: Ah that's great!

PR #320, opened 2021-03-29
fix: Don't generate metadata when there are no services

jskeet:
(Example: Google.Cloud.OsLogin.Common, where we just generate resources names.)

Ideally we'd actually delete any existing metadata file, but we don't have a way of expressing that - and it's very unlikely to happen.


viacheslav-rostovtsev (approved):
so the comment on the line 203 seems to be misleading. 
`// When processing a (proto) package without any services there will be no generated content.`

jskeet:
Yup - it really means "without any services or resource names".

jskeet:
I'll wait until tomorrow to merge https://github.com/googleapis/google-cloud-dotnet/pull/6286, and then regenerate manually in one big PR.

PR #322, opened 2021-03-30
fix: Make metadata generation order stable

jskeet:
(It would be nice if `MapField` contains an `Add` overload accepting an
`IEnumerable<TKey, TValue>`, but it doesn't at the moment. Using
SortedDictionary is a fairly simple alternative though.)


amanda-tarafa (approved):
LGTM

PR #324, opened 2021-03-30
fix: Clear pending trivia only if it was attached to a statement.

amanda-tarafa:
As a side effect of fixing this bug, we now need to be more explicit in whether we include some trivia or not, which is a good thing. This has only affected snippet generation, but won't have any effect on generated code.

This PR shouldn't have any effect on generated code, thus it shouldn't trigger regeneration. I'm checking that locally though. In the meantime, I'll add the `do not merge` label. But this is ready for review.


jskeet (approved):

amanda-tarafa:
I've confirmed that this won't trigger regeneration.
Merging.

PR #323, opened 2021-03-30
feat: Remove doc markers from standalone snippets.

amanda-tarafa:
- First commit is a simple fix to an infrequent bug, catched while writing the code for the feature described in this PR.
- Second commit contains the actual feature described in this PR.
  - Production code changes are only in SnippetGenerator.cs. The rest is just the changes to the expected generator output because of this change.
- This PR shouldn't trigger regenation, but I'm running it locally to confirm that. In the meantime, I've added the `do not merge` label. But this is ready for review.

PR #325, opened 2021-03-30
feat: Remove doc markers from standalone snippets.

amanda-tarafa:
Standalone snippets will use region tags and metadata.

- Production code changes in SnippetCodeGenerator.cs only. The rest is just changes to the expected output.
- This shouldn't trigger regeneration but as usual I'm running it on local on won't merge until that's done.
- @jskeet as for your recommendation from #323 review to use a named parameter, that's a Func, so I can't. I've added a comment though.

### DIFF
--- a/Google.Api.Generator/Generation/SnippetCodeGenerator.cs
+++ b/Google.Api.Generator/Generation/SnippetCodeGenerator.cs
@@ -142,13 +142,14 @@ namespace Google.Api.Generator.Generation
             private bool Sync { get; }
             public string SnippetMethodName { get; }
             private string TargetMethodName => Sync ? TargetMethod.SyncMethodName : TargetMethod.AsyncMethodName;
-            private Func<SourceFileContext, MethodDeclarationSyntax> MethodGenerator { get; }
+            private Func<SourceFileContext, bool, MethodDeclarationSyntax> MethodGenerator { get; }
 
-            public SnippetDef(MethodDetails targetMethod, bool sync, string snippetMethodName, Func<SourceFileContext, MethodDeclarationSyntax> methodGenerator) =>
+            public SnippetDef(MethodDetails targetMethod, bool sync, string snippetMethodName, Func<SourceFileContext, bool, MethodDeclarationSyntax> methodGenerator) =>
                 (TargetMethod, Sync, SnippetMethodName, MethodGenerator) = (targetMethod, sync, snippetMethodName, methodGenerator);
 
-            public MethodDeclarationSyntax GenerateMethod(SourceFileContext ctx) => 
-                MethodGenerator(ctx)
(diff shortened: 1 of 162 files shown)


Google.Api.Generator/Generation/SnippetCodeGenerator.cs
```diff
@@ -327,32 +333,32 @@ private IEnumerable<ObjectInitExpr> InitRequest()
             private MethodDeclarationSyntax Sync(string methodName, IEnumerable<Typ> snippetTyps, object initRequest, object makeRequest) =>
                 Method(Public, VoidType, methodName)()
                     .WithBody(
-                        $"// Snippet: {Method.SyncMethodName}({SnippetTypes(snippetTyps)}{nameof(CallSettings)})",
+                        IncludeDocMarkers ? $"// Snippet: {Method.SyncMethodName}({SnippetTypes(snippetTyps)}{nameof(CallSettings)})" : null,
```
jskeet, line 336: Just as a thought, perhaps a method of `DocMarker(string value) => IncludeDocMarker ? value : null;` would simplify this code a bit? Could definitely be done as a follow-up though. (Just having the same logic many, many times bothers me a bit. I can see us having to change them all again, where with a suitable abstraction we could change one thing.)

Google.Api.Generator/Generation/SnippetCodeGenerator.cs
```diff
@@ -699,38 +728,45 @@ public static IEnumerable<ResourceName> Create(Signature sig)
                         (object)_sig._def.Client.Call(Method.SyncMethodName)(InitRequestArgs.ToArray()) :
                         _sig._def.Response.WithInitializer(_sig._def.Client.Call(Method.SyncMethodName)(InitRequestArgs.ToArray())));
 
-                    public SnippetDef SyncSnippet => new SnippetDef(Method, true, SyncMethodName, ctx => WithSourceFileContext(ctx).SyncMethod);
+                    public SnippetDef SyncSnippet => new SnippetDef(Method, true, SyncMethodName,
```
jskeet, line 731: Likewise I suspect another abstraction here could simplify things... but I suspect I'd have to create a separate PR to try it.

jskeet (approved):
Nit: 2 commits would still have been easier to review :)

As noted in comments, I'd like to remove some of this repetition, but I suspect it will be easier for me to just try it in a separate PR than describe it in detail now, if you're okay with that.

amanda-tarafa:
Ack that is better to have the snippet files in a separate commit. I'll do that from now on.
My next PR is just refactoring, including around some of the suggestions you made. I just wanted to get the snippet files out of the way first with the change looking more staightforward. Hopefully it will address your concerns.

I've confirmed that this won't trigger regeneration so I'll merge now.

PR #326, opened 2021-04-01
fix(REST): Type and property declaration for keyword named enums didn't match.

amanda-tarafa:
Found when generating https://admin.googleapis.com/$discovery/rest?version=directory_v1, which contains an enum named `event`.


amanda-tarafa:
@jskeet I'm pretty sure we need this change, but I've just tested if with Admin Directory. In any case I wouldn't merge before Tuesday, so I've created a draft PR.

jskeet:
Will look first thing on Tuesday morning.

jskeet (approved):
I'm fine in general - have you rerun against *all* APIs yet to check it doesn't break anything?

amanda-tarafa:
I've run against all APIs now without any changes. I've also compared the Admin Directory library generated with this change in with the last Admin Directory library that we released (which was generated with the old python generator). All changes there are true API changes.
I'll merge now.

PR #327, opened 2021-04-02
feat: implement the mixin service generation

viacheslav-rostovtsev:


Google.Api.Generator/Google.Api.Generator.csproj
```diff
@@ -15,6 +15,7 @@
     <PackageReference Include="Google.LongRunning" Version="2.0.0-beta01" />
     <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="2.10.0" />
     <PackageReference Include="Moq" Version="4.12.0" />
+    <PackageReference Include="SharpYaml" Version="1.6.6" />
```
jskeet, line 18: Is there a reason for picking SharpYaml here? We're already using YamlDotNet elsewhere, and I think it would be better to use a single implementation across multiple tools if possible. That said, if SharpYaml has features that YamlDotNet doesn't (and that you need) that's fine.

Google.Api.Generator/Generation/ServiceYaml.cs
```diff
@@ -0,0 +1,45 @@
+﻿using System;
```
jskeet, line 1: Copyright

Google.Api.Generator/Generation/ServiceYaml.cs
```diff
@@ -0,0 +1,45 @@
...
+
+namespace Google.Api.Generator.Generation
+{
+    public class ServiceYaml
+    {
+        public static Service ParseYaml(string yamlContentRaw)
+        {
+            var yamlContentTrim = yamlContentRaw.Trim();
+
+            // If the yaml starts with the `type` attribute, it should be discarded.
```
jskeet, line 17: I would clarify that only the type attribute should be discarded... I misread this to start with as "just ignore the whole YAML file".

jskeet: (It's not clear to me why we need to discard it, to be honest... shouldn't we validate that it has the expected value?)

Google.Api.Generator/Generation/ServiceYaml.cs
```diff
@@ -0,0 +1,45 @@
...
+                }
+
+                yamlContentTrim = yamlContentSplit[1];
+            }
+                
+            var deSerializer = new Serializer();
+            var yamlParseResult = deSerializer.Deserialize(yamlContentTrim);
+
+            var serializer = new Serializer(new SerializerSettings { EmitJsonComptible = true });
+            var yaml2JsonRaw = JsonConvert.SerializeObject(yamlParseResult, Formatting.Indented);// serializer.Serialize(yamlParseResult);
```
jskeet, line 35: Remove comment?

Google.Api.Generator.Tests/YamlParserTest.cs
```diff
@@ -0,0 +1,228 @@
...
+{
+    public class YamlParserTest
+    {
+        [Fact]
+        public void CheckDeserialization()
+        {
+            const string name = "";
+            const string title = "Dialogflow API";
+            var serviceYaml = String.Format(serviceYamlTemplate, name, title);
+            var syaml = ServiceYaml.ParseYaml(serviceYaml);
```
jskeet, line 32: Possibly renamed this variable to "service"?

jskeet:
Should we close this in favour of #406?

PR #329, opened 2021-04-07
fix: Use Google LLC as the copyright and author name

jskeet:
I'm regenerating the REST libraries with this at the moment. Will only merge when both PRs have been approved.


amanda-tarafa (approved):
LGTM.

PR #330, opened 2021-04-13
fix: Add WithGoogleRequestParam for server streaming calls

### LINKED ISSUE #328: x-goog-request-params header not generated for streaming calls
Spotted in https://github.com/googleapis/google-cloud-dotnet/issues/6310

In [ServiceImplClientClassGenerator](https://github.com/googleapis/gapic-generator-csharp/blob/master/Google.Api.Generator/Generation/ServiceImplClientClassGenerator.cs#L116) we don't generate calls to WithGoogleRequestParam for streaming calls (either bidi or server streaming).

It's slightly surprising that this hasn't caused any problems before, and we should be careful to check that it won't break anything if we *do* start generating it everywhere, but it looks like it would at least be simple to fix if we *do* want to generate that everywhere.

jskeet:
Fixes #328


jskeet:
Note: The "do not merge" label is so that I can run the generator with this change on google-cloud-dotnet and check all the output, including running any integration tests.

viacheslav-rostovtsev (approved):

jskeet:
All tests passing with new code - merging.

PR #334, opened 2021-05-04
Throw exception on invalid resource annotation

### LINKED ISSUE #333: Feature request: fail generation if resource name annotations are on non-string fields
https://github.com/googleapis/google-cloud-dotnet/issues/6425 is caused by generating invalid code based on invalid proto configuration. If the generator failed in a really obvious way, that would be better than generating invalid code (and would have a chance at avoiding the broken protos from being checked in to start with).

I'll have a look and see if I can implement this myself...

jskeet:
If a resource_reference annotation is specified for a field of any
type other than string, we should throw an exception rather than
generating invalid code. This should allow presubmits to fail,
avoiding broken protos to be checked in.

Fixes #333.

Tested by:

- Demonstrating that the current Dialogflow protos fail generation (see below)
- Generating all (currently-generated-for-C#) APIs after fixing Dialogflow: no changes to current results

Failure with current Dialogflow proto:

>  System.InvalidOperationException: Field UpdateEnvironmentRequest.environment has a resource reference annotation, but is not a string field

### DIFF
--- a/Google.Api.Generator/ProtoUtils/ResourceDetails.cs
+++ b/Google.Api.Generator/ProtoUtils/ResourceDetails.cs
@@ -15,6 +15,7 @@
 using Google.Api.Gax;
 using Google.Api.Generator.Utils;
 using Google.Protobuf.Reflection;
+using Google.Protobuf.WellKnownTypes;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
 using System;
 using System.Collections.Generic;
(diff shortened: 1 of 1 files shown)


jskeet:
Test failure is due (I believe) to one of the test protos being incorrectly specified. Will fix that now.

jskeet:
Ah - no, it's to support StringValue as well. Will fix.

jskeet:
Fixed now. (Not sure why we can't compare message descriptors directly, but not to worry - comparing the full name is fine.)

alexander-fenster (approved):

PR #338, opened 2021-05-19
Disable deprecation warnings in snippets calling PollOnce for deprecated LROs

jskeet:
(No test as it would take massively longer to add a test than to
just check. The real test is that the API that is currently broken
is fixed by this change.)


jskeet:
The change is reflected in the fourth commit in https://github.com/googleapis/google-cloud-dotnet/pull/6477

amanda-tarafa (approved):
LGTM

PR #344, opened 2021-07-20
feat: pagination detection/generation for Diregapic

viacheslav-rostovtsev:


Google.Api.Generator/Generation/MethodDetails.cs
```diff
@@ -275,6 +275,72 @@ private static MethodDetails DetectPagination(ServiceDetails svc, MethodDescript
...
+            return null;
+
+            bool IsResultsLimitCandidate(FieldDescriptor field) =>
+                field is {FieldType: FieldType.Int32, IsRepeated: false};
+
+            bool IsRepeatedCandidate(FieldDescriptor field) =>
+                field.IsRepeated && !field.IsMap && (field.FieldType == FieldType.Message || field.FieldType == FieldType.String);
+        }
+
+        private static FieldDescriptor FindResultsLimitCandidate(MessageDescriptor input, Func<FieldDescriptor, bool> isResultsLimitCandidate)
```
jskeet, line 318: Rather than pass this as a delegate, could we just make `IsResultsLimitCandidate` a regular method and call it directly from this method? I think it would make things simpler. (I started looking to see whether we used different methods for this parameter.)

Google.Api.Generator/Generation/MethodDetails.cs
```diff
@@ -275,6 +275,72 @@ private static MethodDetails DetectPagination(ServiceDetails svc, MethodDescript
...
+            var input = desc.InputType;
+            var output = desc.OutputType;
+
+            var resultsLimitCandidate = FindResultsLimitCandidate(input, IsResultsLimitCandidate);
+            var pageTokenCandidate = input.FindFieldByName("page_token");
+            var nextPageTokenCandidate = output.FindFieldByName("next_page_token");
+
+            var mapCandidateFields = output.Fields.InDeclarationOrder().Where(f => f.IsMap).ToList();
+            var repeatedCandidateFields = output.Fields.InDeclarationOrder().Where(IsRepeatedCandidate).ToList();
+            var repeatedField = mapCandidateFields.SingleOrDefault() ?? repeatedCandidateFields.SingleOrDefault();
```
jskeet, line 289: If there are multiple map fields or multiple repeated fields, I believe it's still valid so long as the first one numerically is also the first one in declaration order. At least, that's what the AIP says for regular pagination.

Also, if there *are* multiple fields, SingleOrDefault will give an exception which isn't terribly descriptive - I think it might be better to be explicit about how we handle multiple fields.

jskeet (commented):
Some initial thoughts. Is there an expectation that we'll document the requirements carefully in the AIP?

Google.Api.Generator/Generation/MethodDetails.cs
```diff
@@ -275,6 +275,72 @@ private static MethodDetails DetectPagination(ServiceDetails svc, MethodDescript
...
+        private static MethodDetails DetectRegapicPagination(ServiceDetails svc, MethodDescriptor desc)
+        {
+            var input = desc.InputType;
+            var output = desc.OutputType;
+
+            var resultsLimitCandidate = FindResultsLimitCandidate(input, IsResultsLimitCandidate);
+            var pageTokenCandidate = input.FindFieldByName("page_token");
+            var nextPageTokenCandidate = output.FindFieldByName("next_page_token");
+
+            var mapCandidateFields = output.Fields.InDeclarationOrder().Where(f => f.IsMap).ToList();
```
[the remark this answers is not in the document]
viacheslav-rostovtsev: I know that a `proto3_optional` is a `oneof` under the hood, if that is what you are referring to. I also know that a `map` is a `repeated` under the hood, but I don't know of anything that ends up as a `map`.

Google.Api.Generator/Generation/MethodDetails.cs
```diff
@@ -275,6 +275,72 @@ private static MethodDetails DetectPagination(ServiceDetails svc, MethodDescript
                 field.IsRepeated && !field.IsMap && (field.FieldType == FieldType.Message || field.FieldType == FieldType.String);
         }
 
+        private static MethodDetails DetectRegapicPagination(ServiceDetails svc, MethodDescriptor desc)
+        {
+            var input = desc.InputType;
+            var output = desc.OutputType;
+
+            var resultsLimitCandidate = FindResultsLimitCandidate(input, IsResultsLimitCandidate);
```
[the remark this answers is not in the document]
viacheslav-rostovtsev: 👍

Google.Api.Generator/Generation/MethodDetails.cs
```diff
@@ -275,6 +275,72 @@ private static MethodDetails DetectPagination(ServiceDetails svc, MethodDescript
...
+                field.IsRepeated && !field.IsMap && (field.FieldType == FieldType.Message || field.FieldType == FieldType.String);
+        }
+
+        private static FieldDescriptor FindResultsLimitCandidate(MessageDescriptor input, Func<FieldDescriptor, bool> isResultsLimitCandidate)
+        {
+            // Has the int32 page_size or int32 max_results field which defines the maximum number of paginated resources to return in the response
+            var pageSizeCandidate = input.FindFieldByName("page_size");
+            var maxResultsCandidate = input.FindFieldByName("max_results");
+
+            if (pageSizeCandidate == null && maxResultsCandidate == null)
```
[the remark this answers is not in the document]
viacheslav-rostovtsev: I can get behind `is null` and `is not null` but `is object` seems to me a highly unintuitive way to represent `!= null`.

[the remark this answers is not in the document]
viacheslav-rostovtsev: 👍

Google.Api.Generator/Generation/MethodDetails.cs
viacheslav-rostovtsev: 👍

Google.Api.Generator/Generation/MethodDetails.cs
viacheslav-rostovtsev: 👍

Google.Api.Generator/Generation/MethodDetails.cs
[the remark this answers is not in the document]
jskeet: You get used to `is object` pretty quickly to be honest :)

Google.Api.Generator/Generation/MethodDetails.cs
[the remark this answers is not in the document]
jskeet: Right. We're probably fine - just something to keep an eye on. (I suspect it was proto3 optional fields that I was thinking of, so at least we don't need to worry about that.)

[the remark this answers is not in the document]
viacheslav-rostovtsev: 👍

jskeet:
I'm on vacation until Wednesday - will look then.

Google.Api.Generator.Utils/Roslyn/RoslynBuilder.cs
```diff
@@ -315,9 +315,11 @@ static void AppendString(string s, List<InterpolatedStringContentSyntax> list)
         public static SwitchFunc Switch(PropertyDeclarationSyntax expr) => Switch(IdentifierName(expr.Identifier));
 
         public static CastExpressionSyntax Cast(TypeSyntax type, ExpressionSyntax expr) => CastExpression(type, expr);
-
+
```
jskeet, line 318: Nit: remove this whitespace-only change?

Google.Api.Generator/Generation/MethodDetails.cs
```diff
@@ -253,37 +253,112 @@ private static MethodDetails DetectPagination(ServiceDetails svc, MethodDescript
...
+        /// That is a field that is named "page_size" or "max_results" and has a non-repeated Int32 type, if any.
+        /// If no such field exists, then simply a field that is named "page_size" or "max_results".
+        /// </summary>
+        private static FieldDescriptor FindPageSizeCandidate(MessageDescriptor input)
+        {
+            // Has the int32 page_size or int32 max_results field which defines the maximum number of paginated resources to return in the response
+            var pageSizeCandidate = input.FindFieldByName("page_size");
+            var maxResultsCandidate = input.FindFieldByName("max_results");
+
+            if (pageSizeCandidate is null && maxResultsCandidate is null)
```
jskeet, line 336: I'd be tempted to do all of this with a conditional operator:

```csharp
return pageSizeCandidate is null && maxResultsCandidate is null ? null
  : IsValidPageSizeCandidate(pageSizeCandidate) ? pageSizeCandidate
  : IsValidPageSizeCandidate(maxResultsCandidate) ? maxResultsCandidate
  : pageSizeCandidate ?? maxResultsCandidate;
```

But I'm more of a fan of the conditional operator than some other folks. No need to change, just something to consider.

Google.Api.Generator/Generation/MethodDetails.cs
```diff
@@ -253,37 +253,112 @@ private static MethodDetails DetectPagination(ServiceDetails svc, MethodDescript
...
+            // returning the non-null candidate here even though it did not pass all the requirements
+            // an exception should be thrown but only if method has _all_ candidates, which we will only know later.
+            return pageSizeCandidate ?? maxResultsCandidate; 
+        }       
+
+        /// <summary>
+        /// The non-name criteria for a field to be a suitable candidate for a `page size`/`max results` request field.
+        /// </summary>
+        private static bool IsValidPageSizeCandidate(FieldDescriptor field) =>
+            field!=null && ((field.FieldType == FieldType.Int32 || field.FieldType == FieldType.UInt32) && !field.IsRepeated);
```
jskeet, line 360: Nit: `field!=null` => `field != null` (or `field is object`)

Google.Api.Generator/Generation/MethodDetails.cs
jskeet: Also note that we don't actually need the first line at all - if they're both null, the rest of the conditions will end up returning null.

Google.Api.Generator/Generation/ServiceCodeGenerator.cs
```diff
@@ -78,6 +82,37 @@ private static IEnumerable<MemberDeclarationSyntax> PaginatedPartialClasses(Sour
...
+
+                using (ctx.InClass(partialInterfaceCls))
+                {
+                    var underlyingProperty = Property(DontCare, ctx.TypeDontCare, "MaxResults");
+
+                    var assignmentVal = ProtoTyp.Of(maxResMessage) == Typ.Of<int>() 
+                        ? Value
+                        : CheckedCast(ctx.Type(ProtoTyp.Of(maxResMessage)), Value);
+
+                    var property =  Property(Public, ctx.Type<Int32>(), "PageSize")
```
jskeet, line 106: Have you considered making this an explicitly-implemented property instead of a public one? I think that might be less confusing, if it works.

Google.Api.Generator/ProtoUtils/ProtoTyp.cs
```diff
@@ -83,12 +83,16 @@ public static Typ Of(FieldDescriptor desc, bool? forceRepeated = null)
         {
             if (desc.IsMap)
             {
+                var kv = desc.MessageType.Fields.InFieldNumberOrder();
                 if (forceRepeated != null)
                 {
-                    throw new InvalidOperationException("Cannot force repeated on a map field.");
+                    // it turns out that you can force repeated on a map field.
```
jskeet, line 89: Nit: please start comment sentences ("it" and "this") with a capital letter. This is a personal bugbear - hope you're okay changing it. (If you feel passionately in the opposite direction, feel free to leave it though...)

jskeet (approved):
A few nits, but that's all.

Google.Api.Generator.Utils/Roslyn/RoslynBuilder.cs
viacheslav-rostovtsev: 👍

Google.Api.Generator/Generation/MethodDetails.cs
viacheslav-rostovtsev: 👍 
also, my dissatisfaction about `is object` syntax, through the lens of C# documentation:
"The is operator checks if the result of an expression is compatible with a given type" (https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/is) 
why do we check that `FieldDescriptor` is compatible with `Object` again?
Oh, it's because the full specification clarifies that the null check is performed
"If E is a method group or the null literal, of if the type of E is a reference type or a nullable type and the value of E is null, the result is false."
(https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/expressions#the-is-operator)
at this point it's like the question of 'why are we hitting ourselves', luckily the C# 9 will give us `is not null` and make this moot.

Google.Api.Generator/ProtoUtils/ProtoTyp.cs
viacheslav-rostovtsev: 👍

Google.Api.Generator/Generation/ServiceCodeGenerator.cs
viacheslav-rostovtsev: Turns out this does not quite work because we do generate a code accessing the public setter, e.g.:
```
        public virtual gax::PagedEnumerable<AddressAggregatedList, scg::KeyValuePair<string, AddressesScopedList>> AggregatedList(string project, string pageToken = null, int? pageSize = null, gaxgrpc::CallSettings callSettings = null) =>
            AggregatedList(new AggregatedListAddressesRequest
            {
                Project = gax::GaxPreconditions.CheckNotNullOrEmpty(project, nameof(project)),
                PageToken = pageToken ?? "",
                PageSize = pageSize ?? 0,
            }, callSettings);
```

I have followed your suggestions in the Compute PR and have added code for generating a public getter alongside the setter.

Google.Api.Generator/Generation/MethodDetails.cs
viacheslav-rostovtsev: 👍

Google.Api.Generator/Generation/ServiceCodeGenerator.cs
jskeet: Ah, that's a pity. (And now you mention it, that makes sense as I think I tried to do it myself.)

We could *potentially* set MaxResults instead, but making PageSize public instead is okay.

Google.Api.Generator.Utils/Roslyn/RoslynBuilder.cs
```diff
@@ -233,6 +233,8 @@ public static PropertyDeclarationSyntax AutoProperty(Modifier modifiers, TypeSyn
         public static CodeFunc<ForStatementSyntax> For(LocalDeclarationStatementSyntax decl, ExpressionSyntax condition, ExpressionSyntax incr) => code =>
             ForStatement(decl.Declaration, SeparatedList<ExpressionSyntax>(), condition, SingletonSeparatedList(incr), Block(ToStatements(code)));
 
+        public static ExpressionSyntax ToDefaultExpression(object foo) => ToExpression(foo);
```
jskeet, line 236: Not sure what this new method is needed for? (Or what it does.) Let's remove it if we don't need it, but I'm fine if we do.

jskeet (approved):
Righto - one question, but otherwise ready for you to rebase or squash to a nicer commit set, and merge.

(I note there's no "golden" test case for this. It's probably worth adding one later, but I don't think it needs to block this.)

amanda-tarafa (approved):
LGTM

Google.Api.Generator.Utils/Roslyn/RoslynBuilder.cs
viacheslav-rostovtsev: 👍

PR #346, opened 2021-07-27
fix: Generate appropriate XML docs for generic paginated resources

jskeet:
In XML docs, cref attributes can specify open generic types, but not
constructed ones... which means we can't just use the paginated
resource type directly in all cases.

This is a potentially more general problem, but the fact that it
hasn't bitten us elsewhere is encouraging - if it strikes anywhere
else, we'll know because the generated code won't compile cleanly.


jskeet:
Looking at the test failure now...

amanda-tarafa (approved):
LGTM.

jskeet:
(Will wait for Slava to look before merging.)

viacheslav-rostovtsev (approved):

viacheslav-rostovtsev:
LGTM!

PR #364, opened 2021-07-30
fix: make bazel examples work

viacheslav-rostovtsev:


vam-google (approved):
LGTM, though ideally we want protobuf version match googleapis version.

jskeet (approved):

PR #365, opened 2021-08-04
chore: Add System.Collections.Immutable dependency

jskeet:
It's unclear exactly why this is required, but it may well be due to
a separate recent dependency update. With this in place, it's fine.


amanda-tarafa (approved):
LGTM

PR #368, opened 2021-08-06
Support for repeated enum query parameters

jskeet:


Google.Api.Generator.Rest/Models/ParameterModel.cs
```diff
@@ -91,6 +97,21 @@ private PropertyDeclarationSyntax GenerateProperty(SourceFileContext ctx)
                 .WithAttribute(ctx.Type<RequestParameterAttribute>())(Name, locationExpression);
             if (_schema.Description is string description)
             {
+                // TODO: Add a remarks section for repeated enum parameters.
+                property = property.WithXmlDoc(XmlDoc.Summary(description));
+            }
+            return property;
+        }
+
+        private PropertyDeclarationSyntax GenerateRepeatedProperty(SourceFileContext ctx)
```
amanda-tarafa, line 106: Wondering if you considered modifying `SchemaTypes.GetTypFromSchema`, so that it returns the `Repeatable<TEnum>` instead of the `TEnum`, and the exception is made here for the single property instead of the repeated one. Just maybe it's a little more realistic as to what is the real supported scenario, and what is the one we are supporting for compatibility.

I'm fine either way though, whatever has the less ramifications works best.

jskeet: Possibly - I'd have to investigate what that does elsewhere, e.g. in resources. It may well be that it just doesn't come up.

Google.Api.Generator.Rest/Models/ParameterModel.cs
```diff
@@ -91,14 +103,37 @@ private PropertyDeclarationSyntax GenerateProperty(SourceFileContext ctx)
...
+        {
+            var propertyType = ctx.Type(Typ.Generic(Typ.Of(typeof(Repeatable<>)), Typ.GenericArgTyps.Single()));
+            var locationExpression = ctx.Type<RequestParameterType>().Access(Location.ToString());
+            var property = AutoProperty(Modifier.Public | Modifier.Virtual, propertyType, PropertyName + "List", hasSetter: true)
+                .WithAttribute(ctx.Type<RequestParameterAttribute>())(Name, locationExpression);
+            if (_schema.Description is string description)
+            {
+                property = property.WithXmlDoc(
+                    XmlDoc.Summary(description),
+                    XmlDoc.Remarks($"Use this property to one or more values for the parameter. Do not set both this property and ", IdentifierName(PropertyName), "."));
```
amanda-tarafa, line 125:
```suggestion
                    XmlDoc.Remarks($"Use this property to set one or more values for the parameter. Do not set both this property and ", IdentifierName(PropertyName), "."));
```

amanda-tarafa (approved):
LGTM, tiny nit

Google.Api.Generator.Rest/Models/ParameterModel.cs
jskeet: Doh, yes - thanks, will fix.

PR #374, opened 2021-08-16
Self-signed JWT opt-in code

jskeet:
As ever, review one commit at a time

### DIFF
--- a/Google.Api.Generator/Generation/ServiceAbstractClientClassCodeGenerator.cs
+++ b/Google.Api.Generator/Generation/ServiceAbstractClientClassCodeGenerator.cs
@@ -14,10 +14,11 @@
 
 using Google.Api.Gax;
 using Google.Api.Gax.Grpc;
-using Google.Api.Generator.Utils.Roslyn;
 using Google.Api.Generator.Utils;
+using Google.Api.Generator.Utils.Roslyn;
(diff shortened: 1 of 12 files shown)


Google.Api.Generator/Generation/ServiceAbstractClientClassCodeGenerator.cs
```diff
@@ -82,7 +85,22 @@ private ClassDeclarationSyntax Generate()
 
         private PropertyDeclarationSyntax ChannelPool(PropertyDeclarationSyntax defaultScopes) =>
```
amanda-tarafa, line 86: Maybe have ChannelPool method receive `PropertyDeclarationSyntax useJwtAccessWithScopes` the same as it receives the `defaultScopes`. Just for simetry. And then just swat the order of lines 55/56.

amanda-tarafa (approved):
LGTM with one suggestion.

Google.Api.Generator/Generation/ServiceAbstractClientClassCodeGenerator.cs
jskeet: Can do. I'll do the same with other things where I'm just creating extra declarations for the heck of it :)

PR #385, opened 2021-09-14
fix: Fix Asset generation which otherwise fails due to unknown resources

jskeet:
Before this commit, generation fails if any resource name field is
loaded without the resource definition being present. That's only
actually a problem if the field is used for generation - otherwise,
it's fine.

This commit fixes Google.Cloud.Asset.V1 to avoid an
otherwise-unnecessary import being required.

Once this has been submitted, we can create a release which will fix the Kokoro builds for googleapis/googleapis, and we can remove the workaround in google-cloud-dotnet.


amanda-tarafa (approved):
LGTM

jskeet:
Thanks Amanda - I'll wait for @viacheslav-rostovtsev to review as well, given that this is a sensitive area.

jskeet:
Merging now to unblock internal CI.

PR #388, opened 2021-09-18
fix: carve a pagination detection exception for bigquery

viacheslav-rostovtsev:
BigQuery libraries do have generation Bazel rules, even though the published library is a handwritten veneer.
Turns out those libraries also have paginated methods with `max_results` and a type google.protobuf.UInt32Value and our new pagination detection falsely identifies those as errors.

### DIFF
--- a/Google.Api.Generator/Generation/MethodDetails.cs
+++ b/Google.Api.Generator/Generation/MethodDetails.cs
@@ -24,7 +24,6 @@ using Grpc.ServiceConfig;
 using System;
 using System.Collections.Generic;
 using System.Collections.Immutable;
-using System.Diagnostics;
 using System.Linq;
 using System.Net;
 using System.Text;
--- a/Google.Api.Generator/Generation/MethodDetails.cs
+++ b/Google.Api.Generator/Generation/MethodDetails.cs
@@ -269,7 +269,7 @@ namespace Google.Api.Generator.Generation
                 return null;
             }
 
-            if (pageSizeCandidate.Name == "max_results" && !(pageSizeCandidate.MessageType is null) && pageSizeCandidate.MessageType.FullName == "google.protobuf.UInt32Value")
+            if (pageSizeCandidate.Name == "max_results" && pageSizeCandidate.MessageType?.FullName == "google.protobuf.UInt32Value")
             {
                 // This happens in BigQuery APIs that should still be generated, without pagination
                 return null;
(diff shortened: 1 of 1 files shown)


Google.Api.Generator/Generation/MethodDetails.cs
```diff
@@ -269,6 +269,12 @@ private static MethodDetails DetectPagination(ServiceDetails svc, MethodDescript
                 return null;
             }
 
+            if (pageSizeCandidate.Name == "max_results" && pageSizeCandidate.MessageType.FullName == "google.protobuf.UInt32Value")
```
jskeet, line 272: `pageSizeCandidate.MessageType` will return null if `pageSizeCandidate` isn't a message-oriented field.

But I don't see how this made it through `FindPageSizeCandidate` in the first place, as that uses IsValidPageSizeCandidate, which should return null because the field isn't an Int32 field or a UInt32 field.

I would have thought we'd *want* to use this field though - it's still a page size field really. We just need to work out how to do so properly, right? So is this just a temporary change until we work out how to use the field properly?

viacheslav-rostovtsev: :+1: , fixed the nullref.

- If FindPageSizeCandidate cannot find the candidate that passes all criteria it returns one that has just the correct name. Then if all other candidates are present we throw. This is what happens here. 
- It's not a `page_size` field, it's a `max_results` field (this is why this error never happened before). So the error happens for a non-AIP LRO for libraries with handwritten veneers. I don't think updating the generator to handle this case is a priority.

Google.Api.Generator/Generation/MethodDetails.cs
```diff
@@ -269,6 +269,12 @@ private static MethodDetails DetectPagination(ServiceDetails svc, MethodDescript
                 return null;
             }
 
+            if (pageSizeCandidate.Name == "max_results" && !(pageSizeCandidate.MessageType is null) && pageSizeCandidate.MessageType.FullName == "google.protobuf.UInt32Value")
```
jskeet, line 272: This is fine, but a simpler option would be:

```csharp
if (pageSizeCandidate.Name == "max_results" && pageSizeCandidate.MessageType?.FullName == "google.protobuf.UInt32Value")
```

jskeet (approved):
One suggestion for simplification, but otherwise fine.

Google.Api.Generator/Generation/MethodDetails.cs
viacheslav-rostovtsev: :+1:

jskeet:
Humbug:

"System.InvalidOperationException: MessageType is only valid for message or group fields."

Will fix this during the UK daytime. (We just need to check that the field type is a message first.)

PR #389, opened 2021-09-20
feat [bazel]: update and redirect proto and grpc rules

viacheslav-rostovtsev:
bringing our local grpc and proto rules (used by bazel_example) to a better state:
(1) `csharp_proto` and `csharp_grpc` rules are now just aliases to the rules from rules_gapic
(2) updated rules_gapic to the latest version

With this bazel_example now generates `Grpc.g.cs` files


jskeet (approved):
I can't say I follow the Bazel directly - and if anyone else can review this, that would be useful. But I don't have any objections.

PR #391, opened 2021-09-22
feat: Include status codes that will be retried within documentation

jskeet:
It's not *completely* clear to me that we want to do this, but I suspect we do. (I'm finding out what other languages do.)

See below for an example of the result of this applied to `Google.Cloud.Vision.V1.ProductServiceSettings`:

Vision sample

### DIFF
--- a/Google.Api.Generator/Generation/ServiceSettingsCodeGenerator.cs
+++ b/Google.Api.Generator/Generation/ServiceSettingsCodeGenerator.cs
@@ -151,7 +151,31 @@ namespace Google.Api.Generator.Generation
                                 $"Retry delay multiplier: {retry.BackoffMultiplier}",
                                 $"Retry maximum delay: {(int)retry.MaxBackoff.TotalMilliseconds} milliseconds.",
                                 $"Maximum attempts: {(retry.MaxAttempts == int.MaxValue ? "Unlimited" : retry.MaxAttempts.ToString())}",
+                                GetRetriableErrorCodeDocs().ToArray(),
                                 timeoutRemark));
+
+                        IEnumerable<object> GetRetriableErrorCodeDocs()
+                        {
+                            var codes = method.MethodRetryStatusCodes.ToList();
+                            if (!codes.Any())
+                            {
+                                // This is a little unusual, in terms of "Here are the retry timing parameters, that will never be used..." but
+                                // I guess the settings could then be augmented with status codes. (And this does happen.)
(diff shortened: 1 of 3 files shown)


Google.Api.Generator/Generation/ServiceSettingsCodeGenerator.cs
```diff
@@ -151,7 +151,31 @@ IEnumerable<PropertyDeclarationSyntax> PerMethod(MethodDetails method)
...
                                 timeoutRemark));
+
+                        IEnumerable<object> GetRetriableErrorCodeDocs()
+                        {
+                            var codes = method.MethodRetryStatusCodes.ToList();
+                            if (!codes.Any())
+                            {
+                                // This is a little unusual, in terms of "Here are the retry timing parameters, that will never be used..." but
+                                // I guess the settings could then be augmented with status codes. (And this does happen.)
+                                yield return "No status codes are retried.";
```
amanda-tarafa, line 164: Can't you get the retry status codes before adding the remark on the method above, and only add the remarks if there are some retriable codes?

amanda-tarafa (approved):
LGTM

Google.Api.Generator/Generation/ServiceSettingsCodeGenerator.cs
jskeet: We could, but my point is that the remarks are still potentially useful - because I think users could fairly easily build new settings based on these ones, *only* changing the status codes, in which case they'd want to know the timing.

amanda-tarafa: Ah yes, I see, agreed.

tbpg:
Should we add a disclaimer that the retryable codes may change in future versions?

tbpg:
Or the retry logic in general?

jskeet:
We don't have any disclaimers on other retry settings - I think I was probably being a little overly cautious when mentioning that in email. (It's easy to end up drowning in disclaimers if we're not careful - they could easily obscure the useful information.)

We *could* add "The current default settings are as follows" to the very start of the remarks, before the list. I'm relatively easy either way.

tbpg:
Sounds good. I'm happy either way.

jskeet:
We've now broadly agreed to do this. Will merge when I've worked out generator versioning concerns. (Mixed OwlBot and Autosynth generation could cause issues.)

PR #396, opened 2021-10-08
Initial prototype of LRO generation for DIREGAPIC

jskeet:
This is used to generate the latest code in
https://github.com/googleapis/google-cloud-dotnet/pull/6952

It's reasonably complete, with the following further work required:

- Generating the partial class for Operation with its ToLroResponse
  method. (That can happen after GA; it's only a single class.)
- When referring to ForwardingCallInvoker, use the GAX version (when
  it exists) - it won't be part of the API.
- Tests! Currently I'm using Compute *as* the test. A smaller test
  API in a new prototest directory would make sense.


Google.Api.Generator.Utils/Roslyn/RoslynBuilder.cs
```diff
@@ -203,6 +203,19 @@ public static PropertyDeclarationSyntax AutoProperty(Modifier modifiers, TypeSyn
             ParameterListSyntax Params() => ParameterList(SeparatedList(Enumerable.Repeat(parameter, parameter == null ? 0 : 1)));
         };
 
+        public static CodeFunc<LambdaExpressionSyntax> Lambda(ParameterSyntax parameter1, ParameterSyntax parameter2) => code =>
```
viacheslav-rostovtsev, line 206: Can this be changed to accommodate any amount of parameters?

Google.Api.Generator/Generation/MethodDetails.cs
```diff
@@ -115,20 +143,40 @@ public Lro(ServiceDetails svc, MethodDescriptor desc) : base(svc, desc)
...
-            public Typ OperationMetadataTyp { get; }
+            public override Typ OperationResponseTyp { get; }
+            public override Typ OperationMetadataTyp { get; }
+        }
+
+        /// <summary>
+        /// Details about an LRO method declared in a Discovery-sourced proto
+        /// using an operation_service annotation. 
+        /// </summary>
+        public sealed class DiscoveryLro : Lro
```
viacheslav-rostovtsev, line 158: I'd rather we not call it DiscoveryLro, because the connection with a discovery document is gone by the time we encounter it. 
 Diregapic or ExtendedLro perhaps?

Google.Api.Generator.Utils/Roslyn/RoslynBuilder.cs
jskeet: Potentially, although I suspect it'll make the code more complicated - and I very much doubt that we'll need more than 2 parameters. (We've managed with just one for a long time.) I'd prefer to wait until there's a compelling reason to generalize before doing so.

Google.Api.Generator/Generation/MethodDetails.cs
jskeet: Diregapic feels more obscure - and still implies Discovery, but also implying REST/JSON. I'm not keen on ExtendedLro either, as that sounds like it's more capable.

How about just NonStandardLro? That's really why we need it - because LROs weren't standardised when these APIs were created, so they don't follow what is now the standard LRO pattern. It also contrasts in an obvious way with StandardLro above :)

Google.Api.Generator.Utils/Roslyn/RoslynBuilder.cs
viacheslav-rostovtsev: I can see someonw writing a 3-param version in the future, but I get your point.

Google.Api.Generator/Generation/MethodDetails.cs
viacheslav-rostovtsev: That does sound good.

jskeet:
@amanda-tarafa: Second commit is ready for review.

I'm currently running this version across *all* APIs in google-cloud-dotnet, expecting it to not change anything except Compute.

PR #400, opened 2021-10-27
feat: Generate gapic_metadata.json by default

jskeet:
This can be disabled using "metadata = False" in the csharp_gapic_library rule.


jskeet:
Closing in favor of #403.

PR #401, opened 2021-10-28
feat: Allow for packaging non-GAPIC libraries

jskeet:
For libraries that contain only proto and/or gRPC components,
without a GAPIC library, we should still be able to create
appropriate output. This requires an extra parameter
(csharp_package) which is usually inferred from the GAPIC client.

### DIFF
--- a/rules_csharp_gapic/csharp_nongapic_pkg.bzl
+++ b/rules_csharp_gapic/csharp_nongapic_pkg.bzl
@@ -1,0 +1,61 @@
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


rules_csharp_gapic/csharp_nongapic_pkg.bzl
```diff
@@ -0,0 +1,61 @@
...
+        "_zipper": attr.label(default = Label("@bazel_tools//tools/zip:zipper"), cfg = "host", executable = True),
+    },
+)
+
+def csharp_nongapic_assembly_pkg(name, deps, csharp_package, **kwargs):
+    _csharp_nongapic_assembly_pkg(
+        name = name,
+        deps = deps,
+        package_dir = name,
+        csharp_package = csharp_package,
```
viacheslav-rostovtsev, line 59: If this parameter is called `package_name` we won't have to whitelist it in build_file_gen as `package_name` is already whitelisted.

viacheslav-rostovtsev (changes requested):

rules_csharp_gapic/csharp_nongapic_pkg.bzl
jskeet: Righto - happy to do that.

jskeet:
Closing this in favor of #404.

PR #403, opened 2021-10-29
feat: Make gapic_metadata.json unconditional

jskeet:
Initially I had expected to only remove the command-line option, but
keep the rest of the code - but it now feels cleaner to get rid of
it all in one go. I suspect if we left the conditionality in, we'd
never actually remove it.


viacheslav-rostovtsev (approved):
LGTM

jskeet:
@amanda-tarafa - it looks like we need your approving review after all.

amanda-tarafa (approved):
LGTM

PR #404, opened 2021-11-02
feat: add generating nongapic rules to csharp_gapic_pkg

viacheslav-rostovtsev:


jskeet (approved):
We could *potentially* get away without `generate_nongapic_package`, using the check of "either gapic_zip is present, or package_name must specified, but not both". But I don't mind being extra-explicit here.

jskeet:
(I won't merge in case you want to remove the `generate_nongapic_package` flag based on my comment.)

PR #405, opened 2021-11-03
feat: generate one gapic_metadata.json for are multiple packages

viacheslav-rostovtsev:


jskeet:
We're discussing this internally.

jskeet (approved):
Discussed offline - happy to merge this for now, even though it will generate unfortunately-too-large gapic_metadata.json files. This unblocks OwlBot migration, and we can use a post-processor temporarily to trim it down appropriately for google-cloud-dotnet.

The long term solution will be to not pass dependencies as "things to generate", but that can wait.

PR #407, opened 2021-11-05
fix: Trim gapic_metadata.json to appropriate services

jskeet:
We assume the first service that was generated has the "right"
protobuf service, and discard any that have different protobuf
packages to that one.

So for example, when generating KMS and IAM, KMS first so only the
KMS services appear in the metadata.

### DIFF
--- a/Google.Api.Generator/Generation/MetadataGenerator.cs
+++ b/Google.Api.Generator/Generation/MetadataGenerator.cs
@@ -32,19 +32,25 @@ namespace Google.Api.Generator.Generation
 
         internal static string GenerateGapicMetadataJson(List<ServiceDetails> allServiceDetails)
         {
+            // We assume that the first service has the correct proto package, and we can discard
+            // services in other packages. This means it doesn't matter (for gapic_metadata.json)
+            // if we're actually asked to generate more services than we really want. This currently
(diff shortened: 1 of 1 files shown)


Google.Api.Generator/Generation/MetadataGenerator.cs
```diff
@@ -32,19 +32,23 @@ internal static class MetadataGenerator
 
         internal static string GenerateGapicMetadataJson(List<ServiceDetails> allServiceDetails)
         {
+            // We assume that the first service has the correct proto package, and we can discard
+            // services in other packages. This means it doesn't matter (for gapic_metadata.json)
```
amanda-tarafa, line 36: nit: Is the last sentence incomplete?

amanda-tarafa (approved):
LGTM

Google.Api.Generator/Generation/MetadataGenerator.cs
jskeet: Whoops, yes. Will fix.

viacheslav-rostovtsev (approved):
I wonder if it exposes us to order instability issues in bazel rules dependencies...
LGTM

jskeet:
If it does, we were already in trouble as we'd be using the wrong proto package.

PR #412, opened 2021-11-16
fix: Use gapic_metadata.json to determine where to copy extra files

jskeet:
This uses the fact that the metadata file is generated from the
first service to be specified to the GAPIC generator, which is
expected to be "the one for the client library".

Other services may be generated (e.g. for mixins) which currently
ends up with files in the wrong place.


amanda-tarafa (approved):
LGTM

rules_csharp_gapic/csharp_gapic_pkg.bzl
```diff
@@ -64,7 +64,7 @@ tar -czhpf {out_tar} -C {out_dir}/.. {pkg_name}
         inputs =  [gapic_zip] + extras
         command = """
 {zipper} x {gapic_zip} -d {out_dir}
-CLIENT_NAME=$(ls -1 {out_dir} | sort | head -n 1)
+CLIENT_NAME=$(grep libraryPackage {out_dir}/gapic_metadata.json | head -n 1 | cut -d'"' -f 4)
```
viacheslav-rostovtsev, line 67: Please add the note from the CL as a comment here for clarity

viacheslav-rostovtsev (approved):
LGTM

rules_csharp_gapic/csharp_gapic_pkg.bzl
jskeet: Added as a new commit for simplicity. I'll squash them into one commit before merging.

PR #422, opened 2022-01-07
chore: Update expected data in tests to specify 2022

jskeet:
Ideally we shouldn't need to do this every year, but let's unblock
other PRs for now...


amanda-tarafa (approved):
LGTM

amanda-tarafa:
There seems to be a couple more missing?

PR #435, opened 2022-02-01
feat: add golden Showcase generation

viacheslav-rostovtsev:


Google.Api.Generator.Tests/ProtoTest.cs
```diff
@@ -72,7 +73,16 @@ public void ProtocExecution()
             // Or include `// TEST_DISABLE` to disable testing of the entire file.
             var dirName = testProtoNames.First();
             var protoPaths = testProtoNames.Select(x => Path.Combine("ProtoTests", dirName, $"{x}.proto"));
-            var files = Run(protoPaths, $"testing.{dirName.ToLowerInvariant()}",
+
+            var package = $"testing.{dirName.ToLowerInvariant()}";
```
jskeet, line 77: I think a comment here would be useful to explain things. Again, I'm happy to do that myself as a follow-up.

jskeet (approved):
I'm assuming any earlier resource name issues are now fixed?

PR #439, opened 2022-02-02
fix: Don't use mixins in the same package that generates them

### LINKED ISSUE #438: Mixin generation fails to take "this package's clients" into account
Currently the IAM and Locations packages fail to build because they try to generate proxy clients for themselves (e.g. `IAMPolicyClient` has an `IAMPolicyClient` property.

Should be easy to fix - I just didn't want to forget about it...

jskeet:
For example, LocationsClient shouldn't have a LocationsClient property

Fixes #438

(No extra tests for this, but local generation works now and didn't before.)


amanda-tarafa (approved):
LGTM

PR #442, opened 2022-02-09
chore: update grpc version in bazel_example

viacheslav-rostovtsev:
old version throws an error with Bazel 5


jskeet (approved):
This makes me *slightly* nervous as this is a later version of gRPC than we're actually using, but I don't think it will cause problems.

PR #443, opened 2022-02-10
feat: Support service config in Bazel scripts

jskeet:
The additional parameter to `csharp_gapic_library` is `service_yaml`
even though the protobuf plugin argument is `service-config` - this is
to be consistent with the Bazel rules for other languages.

The service in example.proto has been renamed to avoid naming
collisions (so that we generate C# code that would actually build.)


jskeet:
(Added Amanda as I believe our settings will require her to review anyway...)

viacheslav-rostovtsev (approved):
LGTM

PR #444, opened 2022-02-11
fix: Better deprecation handling

jskeet:
- Fix one missing piece of decorating a property initializer with
  deprecation suppression
- Change deprecation detection for fields to match protoc: if the
  field isn't deprecated, but the parent message is, it's still
  marked as being obsolete


amanda-tarafa (approved):
LGTM

PR #445, opened 2022-02-11
chore: update dependencies

viacheslav-rostovtsev:


jskeet (changes requested):
Please revert the gRPC changes - GAX 3.7.0 will still depend on 2.41.0, and the generated code should too. We could *potentially* update to 2.42.0, but Grpc.Net.Client still doesn't have a 2.43.0, and I'd rather stick to 2.41.0 unless there's a compelling reason to go to 2.42.0.

Obviously this PR can only be merged after GAX 3.7.0 is released anyway.

jskeet (approved):

PR #456, opened 2022-03-25
fix: Address BadCmdLineInvocation failing oddly

jskeet:
This has always passed in CI and on some developer machines, but not
on all. I don't understand why the behavior was inconsistent, but
it's easy enough to address to make it pass regardless of the
precise nature of the failure.


amanda-tarafa (approved):
LGTM

PR #457, opened 2022-03-25
feat: Stabilize snippet metadata file by sorting on service name.

amanda-tarafa:
I ran the generator locally and most (all?) libraries with multiple clients are affected.

### DIFF
--- a/Google.Api.Generator/Generation/SnippetCodeGenerator.cs
+++ b/Google.Api.Generator/Generation/SnippetCodeGenerator.cs
@@ -54,7 +54,7 @@ namespace Google.Api.Generator.Generation
                         Version = serviceDetails.PackageVersion ?? ""
                     }}
                 },
-                Snippets = { snippets }
+                Snippets = { snippets.OrderBy(snippet => snippet.ClientMethod.Method.Service.ShortName, StringComparer.Ordinal) }
             }.ToFormattedJson();
 
         public static CompilationUnitSyntax Generate(SourceFileContext ctx, ServiceDetails svc)


Google.Api.Generator/Generation/SnippetCodeGenerator.cs
```diff
@@ -54,7 +54,7 @@ internal class SnippetCodeGenerator
                         Version = serviceDetails.PackageVersion ?? ""
                     }}
                 },
-                Snippets = { snippets }
+                Snippets = { snippets.OrderBy(snippet => snippet.ClientMethod.Method.Service.ShortName) }
```
jskeet, line 57: I'd specify `StringComparer.Ordinal` here.

jskeet (approved):
One nit, but otherwise it looks okay as far as it goes - I'll need to check the proto to see if there are other repeated fields to consider.

Google.Api.Generator/Generation/SnippetCodeGenerator.cs
amanda-tarafa: Done

amanda-tarafa (commented):
This is the [snippet_index.proto](https://github.com/googleapis/gapic-generator-csharp/blob/main/Google.Api.Generator/ConfigProtos/snippet_index.proto). There's no other top-level repeated filed except for the one that's being sorted here.
We could sort more after sorting on service name, like RPC name and then even overload name, but none of the other generator bits are sorted like that, so I figured it is best to keep snippet metadata similar to the rest. If we ever want to stabilize more, I think we should do it on a top level basis (sort services and rpcs before generating anything) than on a case by case basis.
After this PR nothing changed if the order of services changes, but everything will change if the order of RPCs within a service changes.
Let me know.

PR #474, opened 2022-05-06
chore: Update xunit.runner.visualstudio to v2.4.5

jskeet:


amanda-tarafa (approved):
LGTM (we test here in netcoreapp3.1 and that works)

PR #475, opened 2022-05-06
feat: Support a "transport" option for generation

jskeet:
This will allow generation to specify when we should allow gRPC/REST/both transports

Note that this is "transport" for consistency with other languages, despite effectively being plural.


jskeet:
Note: we'll need to modify the Bazel rules to allow this to propagate through, but I'll do that separately.

amanda-tarafa (approved):
LGTM

PR #494, opened 2022-06-13
fix: Split transports on "+" instead of ";"

jskeet:
(Corresponding documentation change will be needed in google-cloud-dotnet)


amanda-tarafa (approved):
LGTM

PR #501, opened 2022-07-22
feat: Generator support for REST numeric enums

jskeet:
Note that this should not be *used* until GAX 4.1.0 is released.
(But it would be fine to be present and unused.)

### DIFF
--- a/Google.Api.Generator.Tests/ProtoTests/Showcase/Google.Showcase.V1Beta1/PackageApiMetadata.g.cs
+++ b/Google.Api.Generator.Tests/ProtoTests/Showcase/Google.Showcase.V1Beta1/PackageApiMetadata.g.cs
@@ -25,7 +25,7 @@ namespace Google.Showcase.V1Beta1
     internal static class PackageApiMetadata
     {
         /// <summary>The <see cref="gaxgrpc::ApiMetadata"/> for services in this package.</summary>
-        internal static gaxgrpc::ApiMetadata ApiMetadata { get; } = new gaxgrpc::ApiMetadata("Google.Showcase.V1Beta1", GetFileDescriptors);
+        internal static gaxgrpc::ApiMetadata ApiMetadata { get; } = new gaxgrpc::ApiMetadata("Google.Showcase.V1Beta1", GetFileDescriptors).WithRequestNumericEnumJsonEncoding(true);
 
         private static scg::IEnumerable<gpr::FileDescriptor> GetFileDescriptors()
         {
--- a/Google.Api.Generator/Program.cs
+++ b/Google.Api.Generator/Program.cs
@@ -33,6 +33,7 @@ namespace Google.Api.Generator
         private const string nameServiceConfigYaml = "service-config";
         private const string nameCommonResourcesConfig = "common-resources-config";
         private const string nameTransport = "transport";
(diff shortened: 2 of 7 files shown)


Google.Api.Generator/Program.cs
```diff
@@ -180,9 +184,10 @@ private static int GenerateFromProtoc(Stream stdin, Stream stdout)
                 var serviceConfigPath = extraParams.GetValueOrDefault(nameServiceConfigYaml)?.SingleOrDefault();
                 var commonResourcesConfigPaths = extraParams.GetValueOrDefault(nameCommonResourcesConfig);
                 var transports = ParseTransports(extraParams.GetValueOrDefault(nameTransport)?.SingleOrDefault());
+                var requestNumericEnumJsonEncoding = extraParams.GetValueOrDefault(nameTransport)?.SingleOrDefault() == "true";
```
amanda-tarafa, line 187:
```suggestion
                var requestNumericEnumJsonEncoding = extraParams.GetValueOrDefault(nameRequestNumericEnumJsonEncoding)?.SingleOrDefault() == "true";
```

Google.Api.Generator.Tests/ProtoTests/Showcase/Google.Showcase.V1Beta1/PackageApiMetadata.g.cs
```diff
@@ -25,7 +25,7 @@ namespace Google.Showcase.V1Beta1
     internal static class PackageApiMetadata
     {
         /// <summary>The <see cref="gaxgrpc::ApiMetadata"/> for services in this package.</summary>
-        internal static gaxgrpc::ApiMetadata ApiMetadata { get; } = new gaxgrpc::ApiMetadata("Google.Showcase.V1Beta1", GetFileDescriptors);
+        internal static gaxgrpc::ApiMetadata ApiMetadata { get; } = new gaxgrpc::ApiMetadata("Google.Showcase.V1Beta1", GetFileDescriptors).WithRequestNumericEnumJsonEncoding(true);
```
amanda-tarafa, line 28: Is Showcase run for REST?

Which makes me think, should we fail if ApiMetada has this setting in true, but only supports gRPC?

amanda-tarafa (approved):
LGTM, but there's a copy-paste typo.

Google.Api.Generator/Program.cs
jskeet: Whoops, thanks!

Google.Api.Generator.Tests/ProtoTests/Showcase/Google.Showcase.V1Beta1/PackageApiMetadata.g.cs
jskeet: We don't actually run it at all - but yes, we generate it for REST.
There's a bit of a disconnect though - this is at the ApiMetadata level, but transport support is at the ServiceMetadata level. That may have been a mistake, of course... but we can't easily fix it now, and it makes checking things tricky.

amanda-tarafa: Yes, that makes sense.

jskeet:
Merging now - we can add the check later on if we want to.

PR #503, opened 2022-07-26
feat: Sort metadata file descriptors and extension methods

jskeet:
We're being passed protos in a different order in Bazel vs in our
generation scripts; this change will avoid unnecessary diffs.


jskeet:
The rollout plan would be as normal:

- Merge
- Create a release
- Use the release from google-cloud-dotnet
- Regenerate all APIs and push in one big PR
- Update the WORKSPACE in googleapis/googleapis

jskeet:
(In fact, it's possible that we won't need the "one big PR" - I suspect that *most* APIs are already in alphabetical order, but I'm running a local test to see...)

jskeet:
We *do* need one big PR - but it's not *huge*; it's likely to be about 85 files.

amanda-tarafa (approved):
LGTM

PR #514, opened 2022-08-16
fix: Fix the transport bazel arg (it's not a file)

jskeet:
At the same time, this moves to using opt_args and opt_file_args to
generate cleaner protoc invocations.

### DIFF
--- a/rules_csharp_gapic/csharp_gapic.bzl
+++ b/rules_csharp_gapic/csharp_gapic.bzl
@@ -47,22 +47,24 @@ def csharp_gapic_library(
         **kwargs):
     # Build zip file of gapic-generator output
     name_srcjar = "{name}_srcjar".format(name = name)
-    plugin_file_args = {}
+    opt_file_args = {}
+    opt_args = []
     if grpc_service_config:
-        plugin_file_args[grpc_service_config] = "grpc-service-config"
+        opt_file_args[grpc_service_config] = "grpc-service-config"
     if common_resources_config:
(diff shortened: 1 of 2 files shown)


jskeet:
Unfortunately the 1.45 release was broken - this one definitely *does* work. (I've tested it against the failing googleapis commit.)

rules_csharp_gapic/csharp_gapic.bzl
```diff
@@ -47,22 +47,24 @@ def csharp_gapic_library(
...
+        opt_file_args[grpc_service_config] = "grpc-service-config"
     if common_resources_config:
-        plugin_file_args[common_resources_config] = "common-resources-config"
+        opt_file_args[common_resources_config] = "common-resources-config"
     if service_yaml:
-        plugin_file_args[service_yaml] = "service-config"
+        opt_file_args[service_yaml] = "service-config"
     if rest_numeric_enums:
-        plugin_file_args[rest_numeric_enums] = "rest-numeric-enums"
+        opt_file_args[rest_numeric_enums] = "rest-numeric-enums"
```
viacheslav-rostovtsev, line 59: Probably also not a file

viacheslav-rostovtsev (approved):
LGTM (but check comments)

rules_csharp_gapic/csharp_gapic.bzl
jskeet: Thanks, will fix in the same way as the transport.

jskeet:
I've now tested this with Compute, including with numeric enums.

PR #515, opened 2022-08-17
Fix to one of the Showcase differences

jskeet:
With this change in place, it's fine to have a service called Xyz which contains an RPC also called Xyz.


viacheslav-rostovtsev (approved):
LGTM

PR #521, opened 2022-08-31
tests: Update test dependencies.

amanda-tarafa:


jskeet (approved):
This would be much simpler if we didn't generate tests ;)

PR #522, opened 2022-08-31
fix: Avoid double escaping for XML comments

jskeet:
The first commit just moves the tests for anything in Google.Api.Generator.Utils into its own project. This made it easier to run tests.

The second commit actually fixes the problem. It's possible it only fixes it in some places, of course - but it does fix it where I noticed the problem...

We may need to also adjust the docs post-processor to unescape `&gt;` as well - `>` isn't escaped by protoc, but is escaped by Roslyn.


jskeet:
Have regenerated all of google-cloud-dotnet with this - it makes a pretty significant difference...

amanda-tarafa (approved):
LGTM

jskeet:
Will create a release for this shortly.

PR #531, opened 2022-09-07
chore: First pass at Showcase integration tests

jskeet:
This is primarily aimed at the data-driven REST compliance tests, but starts building out the Echo tests too, including showing how an abstract test base class can be used to collect common gRPC+REST tests, with concrete subclasses inheriting those tests and adding transport-specific tests.

These tests will not run in GitHub actions, although it's always safe to run the tests - they will all be skipped unless the SHOWCASE_ENDPOINT environment variable is set (typically to http://localhost:7469). In future, we'll add nightly integration tests which stand up a local Showcase server before executing the tests.

### DIFF
--- a/Google.Api.Generator.IntegrationTests/ComplianceRestTest.cs
+++ b/Google.Api.Generator.IntegrationTests/ComplianceRestTest.cs
@@ -1,0 +1,101 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using Google.Api.Gax.Grpc;
+using Google.Protobuf;
+using Google.Showcase.V1Beta1;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Google.Api.Generator.IntegrationTests
+{
+    /// <summary>
+    /// Compliance test for the REST transport.
+    /// </summary>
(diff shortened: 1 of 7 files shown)


Google.Api.Generator.IntegrationTests/ShowcaseTestBase.cs
```diff
@@ -0,0 +1,66 @@
...
+using Grpc.Core;
+using System;
+using Xunit;
+
+namespace Google.Api.Generator.IntegrationTests
+{
+    /// <summary>
+    /// Base class for all Showcase tests.
+    /// </summary>
+    public abstract class ShowcaseTestBase<TClient, TBuilder>
```
amanda-tarafa, line 26: Just curious if you considered this being a base fixture class?
Then there's a rest fixture and a gRPC fixture and more specific fixtures if we needed.

It's fine as is though.

Google.Api.Generator.IntegrationTests/ComplianceRestTest.cs
```diff
@@ -0,0 +1,101 @@
...
+        public static IEnumerable<object[]> Tests =>
+            from grp in s_testFile.Group
+            from rpc in grp.Rpcs
+            from req in grp.Requests
+            select new object[] { new SerializableTest(grp.Name, rpc, req) };
+
+        [SkippableTheory]
+        [InlineData(false)]
+        [InlineData(true)]
+        public void GetEnum(bool unknownEnum)
```
amanda-tarafa, line 41: nit: I think there's something like:
```suggestion
        public void GetEnum([Combinatorial] bool unknownEnum)
```

amanda-tarafa (approved):
LGTM

Google.Api.Generator.IntegrationTests/ShowcaseTestBase.cs
jskeet: We could do that, but I suspect we'd end up with more classes that way, and pointless constructors accepting fixtures just to achieve what we've already achieved here. If we end up needing to do more than create a client though, it would make sense.

Google.Api.Generator.IntegrationTests/ComplianceRestTest.cs
jskeet: That's definitely an option if we also reference the relevant NuGet package - but I personally find the inline data simpler to understand anyway :)

PR #532, opened 2022-09-07
feat: Add snippet segments information in snippet metadata.

amanda-tarafa:


Google.Api.Generator.Tests/ProtoTests/Lro/Testing.Lro.GeneratedSnippets/snippet_metadata_testing.lro.json
```diff
@@ -1,272 +1 @@
...
-            "fullName": "testing.lro.Lro"
-          }
-        }
-      },
-      "origin": "API_DEFINITION"
-    }
-  ]
-}
\ No newline at end of file
+﻿// TEST_DISABLE
```
amanda-tarafa, line 1: Snippet code is not being generated either here, so it really makes no sense to generate metadata, less so with line numbers in, etc.

LRO snippets are tested elsewhere, as well as the snippet metadata.

Google.Api.Generator/Generation/SnippetCodeGenerator.cs
```diff
@@ -149,10 +153,90 @@ private static IEnumerable<SnippetDef> Snippets(SourceFileContext ctx, ServiceDe
...
+            // We know need to format the code to make sure that we extract segment line numbers correctly.
+            var formattedCode = CodeFormatter.Format(ctx.CreateCompilationUnit(snippetCode));
+
+            SnippetSegmentFinder.UpdateMetadataWithSegmentInfo(formattedCode, snippetMetadata);
+
+            // Let's return the string so that the code is not formatted or otherwise changed further.
+            return (formattedCode.ToFullString(), snippetMetadata);
+        }
+
+        private class SnippetSegmentFinder : CSharpSyntaxWalker
```
jskeet, line 168: Add a comment to this? I can sort of work out the point, but it would be good to have it written out.

Google.Api.Generator/Generation/SnippetCodeGenerator.cs
```diff
@@ -149,10 +153,90 @@ private static IEnumerable<SnippetDef> Snippets(SourceFileContext ctx, ServiceDe
         private SnippetDef Snippet { get; }
         public string SnippetMethodName => Snippet.SnippetMethodName;
 
-        public (CompilationUnitSyntax, SnippetMetadata) Generate(SourceFileContext ctx)
+        public (string, SnippetMetadata) Generate(SourceFileContext ctx)
```
jskeet, line 156: I can see why we need to do this - but I don't really like it. It feels like if we change the formatting centrally, we'll need to do it here as well.

I don't have a better solution though.

jskeet (approved):

amanda-tarafa (commented):
All comments addressed on the second commit. PTAL.

jskeet (approved):

amanda-tarafa:
Thanks, I'll rebase and merge.

PR #533, opened 2022-09-07
tools: Add scripts to simplify updating protobufs

jskeet:


amanda-tarafa (approved):
LGTM.

Just some of the new .sh files might need execution permission.

jskeet:
I *believe* all the scripts have the right permission. (In this case it's not as important as at other times, as we'll be running them all manually anyway.)

PR #562, opened 2022-11-07
feat: Add client method parameters to snippet metadata.

amanda-tarafa:


Google.Api.Generator/Generation/SnippetCodeGenerator.cs
```diff
@@ -726,7 +762,34 @@ public class Signature
 
                 private string SyncMethodName => Method.SyncMethodName + OverloadDisambiguation;
                 private string AsyncMethodName => $"{Method.AsyncMethodName.Substring(0, Method.AsyncMethodName.Length - 5)}{OverloadDisambiguation}Async";
-                
+
+                private IEnumerable<ClientMethod.Types.Parameter> CoreClientMethodParameters => _sig.Fields.Select(
+                    field => new ClientMethod.Types.Parameter
```
jskeet, line 767: It might be worth separating out helper methods of `CreateParameter(Type, string)` and `CreateParameter(Typ, string)`.

jskeet (approved):
It's tough to know whether this is all the places that might need changing, but it all looks good to me.

Google.Api.Generator/Generation/SnippetCodeGenerator.cs
amanda-tarafa: Yep, done! Will push again and merge on green. Thanks!

PR #584, opened 2023-01-10
feat: Fail generation for misconfigured REST transport

jskeet:


Google.Api.Generator/CodeGenerator.cs
```diff
@@ -164,6 +164,24 @@ internal static class CodeGenerator
             {
                 throw new InvalidOperationException($"Unhandled APIs in service config: {string.Join(", ", unhandledApis)}");
             }
+
+            ValidateTransports(transports, allServiceDetails);
+        }
+
+        private static void ValidateTransports(ApiTransports transports, List<ServiceDetails> services)
+        {
+            // Note: we only look at the HttpRule options in the proto. While others can be provided
```
amanda-tarafa, line 173: Is it possible for a service with no HTTP bindings to have mixin bindings in service config, and that we could generate the library just to use mixins through it?
I get that even if that can happen, is far fetch, so I'm fine with not supporting that unless there's a real need. But maybe just add it to the note.

amanda-tarafa (approved):
LGTM

Google.Api.Generator/CodeGenerator.cs
jskeet: Interesting scenario. Basically the user would be creating a client with no useful purpose other than exposing the mixins. Happy to call that out in a note rather than trying to handle it. (And the mixins don't actually need overrides if the defaults work.)

jskeet: Note added. Will merge on green.

PR #595, opened 2023-02-06
feat: Include metadata entries when snippets cover more than one client method overload.

amanda-tarafa:


Google.Api.Generator/CodeGenerator.cs
```diff
@@ -258,10 +258,14 @@ private static Service ParseServiceConfigYaml(string path)
                         var snippetCtx = SourceFileContext.CreateUnaliased(
                             clock, s_wellknownNamespaceAliases, s_avoidAliasingNamespaceRegex, packageTyps, maySkipOwnNamespaceImport: false);
                         var (snippetCode, snippetMetadata) = snippetGenerator.Generate(snippetCtx);
-                        snippetMetadata.File = $"{serviceDetails.ClientAbstractTyp.Name}.{snippetGenerator.SnippetMethodName}Snippet.g.cs";
-                        var snippetFile = $"{snippetsPathPrefix}{snippetMetadata.File}";
-                        snippets.Add(snippetMetadata);
-                        yield return new ResultFile(snippetFile, snippetCode);
+                        var snippetFile = $"{serviceDetails.ClientAbstractTyp.Name}.{snippetGenerator.SnippetMethodName}Snippet.g.cs";
+                        var snippetFilePath = $"{snippetsPathPrefix}{snippetFile}";
+                        foreach(var md in snippetMetadata)
```
jskeet, line 263: Nit: space after foreach

Google.Api.Generator/Generation/SnippetCodeGenerator.cs
```diff
@@ -184,55 +184,69 @@ private class SnippetDef
             // TODO: Reconsider this ugly hack when refactoring after removing the service-per-file snippets.
             private bool Canonical => RegionTagDisambiguation is null;
             private IEnumerable<ClientMethod.Types.Parameter> ClientMethodParameters { get; }
+            // May be null, in which case this snippet is for a single client method only.
+            private IEnumerable<ClientMethod.Types.Parameter> AdditionalClientMethodParameters { get; }
```
jskeet, line 188: I don't really understand what this would be. If a method is overloaded, there are multiple *lists* of parameters. In other words, I'd expect this to be a `List<List<Parameter>>` or equivalent.

Or is this *only* for the case where the method just varies by "does it take CallSettings or CancellationToken"?

jskeet (commented):
I haven't reviewed in detail, as I think my first question is probably crucial to me understanding the rest :)

Google.Api.Generator/Generation/SnippetCodeGenerator.cs
jskeet: Okay, that's good to know. I think it's definitely worth adding a comment, along the lines of:

```csharp
// Parameters for a single additional method covered by this snippet.
// (If we ever need to cover more than two overloads in the same snippet, this can become a list of lists.)
```

Google.Api.Generator/Generation/SnippetCodeGenerator.cs
```diff
@@ -184,55 +184,69 @@ private class SnippetDef
...
-                    Language = Language.CSharp,
-                    Canonical = Canonical,
-                    Origin = SnippetMetadata.Types.Origin.ApiDefinition,
-                    ClientMethod = new ClientMethod
+                    return new[] { metadata };
+                }
+
+                return new[] { metadata, GenerateMetadata(AdditionalClientMethodParameters) };
+
+                SnippetMetadata GenerateMetadata(IEnumerable<ClientMethod.Types.Parameter> clientMethodParameters) =>
```
jskeet, line 213: I'm assuming this large-looking change really isn't large at all, it's just a refactoring to a local method.

Google.Api.Generator.Tests/ProtoTests/VoidReturn/Testing.VoidReturn.GeneratedSnippets/snippet_metadata_testing.voidreturn.json
```diff
@@ -105,6 +105,55 @@
         }
       ]
     },
+    {
+      "regionTag": "unknown_generated_VoidReturn_VoidMethod_async",
```
jskeet, line 109: Is the region tag expected to be unique? If so, we'll need some sort of disambiguation.

jskeet (approved):
One question about region tags, and one formatting nit, but otherwise fine. It makes much more sense now I understand the use case :)

amanda-tarafa:
Closing, as we'll go with a different approach.

PR #598, opened 2023-02-15
fix: Handle Boolean values in service configs

### LINKED ISSUE #597: Service config parsing broken
In order to parse service config YAML files, we parse them and serialize them to JSON, then parse the JSON as a proto Service message.

Unfortunately, YamlDotNet serializes Boolean values as strings in JSON, and that's invalid JSON for the proto converter.

As a grotty hack, we can replace `"false"` with `false` and `"true"` with `true` in the JSON. A more direct YAML parse operation would be better though...

jskeet:
Fixes #597.

### DIFF
--- a/Google.Api.Generator/CodeGenerator.cs
+++ b/Google.Api.Generator/CodeGenerator.cs
@@ -195,17 +195,27 @@ namespace Google.Api.Generator
                 return null;
             }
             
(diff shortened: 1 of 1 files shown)


Rishabh-V (approved):
LGTM

PR #601, opened 2023-02-17
chore: Next step of updating to Protobuf 3.22.0

jskeet:
The new package includes the compiler messages, so we can remove these (and generation of them).

This change does *not* update the version of protoc we use, or any dependencies. We need to update GAX and the common protos first, then we can complete the generator work.


Rishabh-V (approved):
LGTM

PR #614, opened 2023-04-06
fix: Filter which publication.library_settings we use by version

jskeet:
The "version" is actually the fully-qualified proto package... which we happen to already have to hand.

(This isn't urgent - please wait until you're back!)

### DIFF
--- a/Google.Api.Generator.Tests/ProtoTests/PublishingSettings/ServiceConfig.yaml
+++ b/Google.Api.Generator.Tests/ProtoTests/PublishingSettings/ServiceConfig.yaml
@@ -10,6 +10,7 @@ apis:
 
 publishing:
   librarySettings:
+  - version: testing.publishingsettings.wrong1
   - version: testing.publishingsettings
     dotnetSettings:
       ignored_resources:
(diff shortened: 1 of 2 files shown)


Google.Api.Generator.Tests/ProtoTests/PublishingSettings/ServiceConfig.yaml
```diff
@@ -10,6 +10,7 @@ apis:
 
 publishing:
   librarySettings:
+  - version: testing.publishingsettings.wrong1
```
amanda-tarafa, line 13: I don't understand this test change. Or maybe I am not very familiar with the schema.

From this change it seems like the same library settings may apply to different versions? Like Beta1, Beta2, etc.?

What I was expecting from the old code comment:

```csharp
// Alternatively, ensure that each *published* service config only contains a single LibrarySettings entry.
```
is to have seen a new `librarySettings` property with different settings and a different version, to make sure we grab the one for this library.

amanda-tarafa: Oh, just me confused by yaml. This is doing exactly what I was expecting.

amanda-tarafa (approved):
LGTM

PR #619, opened 2023-04-17
feat: Generate dependencies for known packages

jskeet:
If we see that the NuGet package containing a mixin is required, even if we're not generating it as as mixin, include the package reference.


jskeet:
This *shouldn't* make any difference at all to google-cloud-dotnet, but I'm checking that now. It should improve the number of APIs that build out of the box in googleapis-gen. (I'm hoping it'll fix about half of the errors.)

jskeet:
Test failure is due to not putting Google.LongRunning first. Fixing that now (by changing the expected project file).

amanda-tarafa (approved):
LGTM

PR #620, opened 2023-04-20
feat: Pin generated snippets namespace to GoogleCSharpSnippets.

amanda-tarafa:
This is closer that how user code would look like in that the their namespace is unlikely to resemble those of the libraries they are using. Snippets will requireless namespace aliasing which will make them cleaner.

### DIFF
--- a/Google.Api.Generator/Generation/ServiceDetails.cs
+++ b/Google.Api.Generator/Generation/ServiceDetails.cs
@@ -50,7 +50,10 @@ namespace Google.Api.Generator.Generation
             ProtoPackage = desc.File.Package;
             PackageVersion = ProtoPackage.Split('.').FirstOrDefault(part => ApiVersionPattern.IsMatch(part));
             DocLines = desc.Declaration.DocLines().ToList();
-            SnippetsNamespace = $"{ns}.Snippets";
+            // For snippets, we use a namespace unrelated to the library as that's likely to be similar to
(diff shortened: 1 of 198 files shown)


amanda-tarafa:
See https://github.com/googleapis/google-cloud-dotnet/pull/10223 for the effects of these changes.

Google.Api.Generator/Generation/ServiceDetails.cs
```diff
@@ -50,7 +50,7 @@ public ServiceDetails(ProtoCatalog catalog, string ns, ServiceDescriptor desc, S
             ProtoPackage = desc.File.Package;
             PackageVersion = ProtoPackage.Split('.').FirstOrDefault(part => ApiVersionPattern.IsMatch(part));
             DocLines = desc.Declaration.DocLines().ToList();
-            SnippetsNamespace = $"{ns}.Snippets";
+            SnippetsNamespace = $"GoogleCSharpSnippets";
```
jskeet, line 53: Nit: no need to be interpolated now. Might be worth a comment though to explain how we've arrived at this.

jskeet (approved):
Hooray for being able to simplify! I'd like to keep going with my "don't generate mixins" work before we do another release though, if that's okay.

Google.Api.Generator/Generation/ServiceDetails.cs
amanda-tarafa: Both done, I'll merge on green, and happy to wait till we have more changes here.

PR #621, opened 2023-04-21
fix: Only generate files for mixins if that's *all* we're generating

jskeet:
There are various reasons why it's hard to pass the "correct" set of protos into the generator; this is a workaround which should at least improve matters.

### DIFF
--- a/Google.Api.Generator/CodeGenerator.cs
+++ b/Google.Api.Generator/CodeGenerator.cs
@@ -86,11 +86,11 @@ namespace Google.Api.Generator
             new Regex(@"^System\.?.*", RegexOptions.Compiled | RegexOptions.CultureInvariant),
         };
 
-        private static readonly IReadOnlyList<string> AllowedAdditionalServices = new List<string>
+        private static readonly IReadOnlyList<ServiceDescriptor> AllowedAdditionalServices = new List<ServiceDescriptor>
         {
-            IAMPolicy.Descriptor.FullName,
-            Locations.Descriptor.FullName,
-            Operations.Descriptor.FullName,
+            IAMPolicy.Descriptor,
+            Locations.Descriptor,
+            Operations.Descriptor,
         }.AsReadOnly();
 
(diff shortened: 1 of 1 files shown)


jskeet:
Tested by running the generator on all Bazel targets in googleapis/googleapis - they all pass, and we only end up with a single copy of each mixin. (There are currently 42 copies of Google.Cloud.Iam.V1 in googleapis-gen, for reference.)

Will run the generation script in google-cloud-dotnet locally as well, having temporarily rebased to *exclude* recent snippet changes (so that those don't confuse things).

Google.Api.Generator/CodeGenerator.cs
```diff
@@ -169,6 +177,29 @@ internal static class CodeGenerator
...
+
+            // Returns the result files we should actually create, ignoring mix-ins unless we're actually
+            // generating the mix-in API.
+            // This is necessary because Bazel passes more files to generate than we really want.
+            // Note that this method may modify filesByProtoPackage - we're not using it after calling this method though.
+            static List<ResultFile> GetResultFilesToCreate(Dictionary<string, List<ResultFile>> filesByProtoPackage)
+            {
+                var mixinPackages = AllowedAdditionalServices.Select(svc => svc.File.Package).ToHashSet();
+
+                // Unless we're *only* generating mixins, remove all mixins.
```
amanda-tarafa, line 190: I guess mixins don't have a Common package now, but they might in the future or something. Maybe just pass to this method the name of the main service, which is already stored on the caller I think.

amanda-tarafa (approved):
LGTM

Google.Api.Generator/CodeGenerator.cs
jskeet: Well, we guess at the name of the main service by finding the first generated service... which is no more definitive than this, really. I very, very much doubt that mixins will start having other dependencies. We can cross that bridge when we come to it.

jskeet:
Finished generating locally from google-cloud-dotnet: no changes. Merging.

PR #630, opened 2023-05-23
fix: Dispose of server/bidi streams in snippets

### LINKED ISSUE #629: Generate using statements for snippets with streaming calls
As per [the docs](https://cloud.google.com/dotnet/docs/reference/help/grpc-streaming), streaming calls should be disposed. Our generated snippets don't do this at the moment. They should use a `using` statement to do so.

jskeet:
Fixes #629.


Rishabh-V (approved):
LGTM ( I am not very familiar with this piece but I can understand what is going on)

PR #654, opened 2023-08-07
chore: Strip tripple backticks from XmlDocs

amanda-tarafa:
Note that what this generates is slightly different from what the current postprocessing workaround generates because this removes the backticks before formatting and the postprocessing workaround removes them after. Otherwise, both produce the same output.

Also, I'm now running GAPIC generation locally, but I expect to see no changes there, backticks would have broken docs for GAPICs as well.


amanda-tarafa:
Confirming that GAPIC generation was no different because of this.

jskeet:
As discussed in person, I'm concerned that this will strip triple backticks from GAPIC generated code as well. And I suspect that putting triple backticks on their own line (rather than removing them entirely) would be better (for Discovery). I'll investigate this.

amanda-tarafa:
Yes, I'm closing this one. I missunderstoon the problem and though that the presence of triple backticks was the problem, instead of them just not being on their own line.

PR #704, opened 2023-11-09
feat: Add DateTimeOffset handling for parameters

### LINKED ISSUE #701: REST: Offer DateTimeOffset conversions for parameters
Request parameters with the "date-time" format (e.g. the `timeMin` parameter in `events.list`) end up generating DateTime properties, but not DateTimeOffset properties. We should handle these in the same way as resource properties.

jskeet:
Fixes #701 

(But see internal doc for further things that need investigating before release.)

### DIFF
--- a/Google.Api.Generator.Rest.Tests/GoldenTestData/Google.Apis.Assuredworkloads.v1/Google.Apis.Assuredworkloads.v1.cs
+++ b/Google.Api.Generator.Rest.Tests/GoldenTestData/Google.Apis.Assuredworkloads.v1/Google.Apis.Assuredworkloads.v1.cs
@@ -1,0 +1,2402 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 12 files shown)


jskeet:
(Looking into the test failures. I could have sworn it passed last time I ran it, but fortunately it's failing locally now in the same way. Hopefully it's a minor problem...)

Google.Api.Generator.Rest.Tests/GoldenTestData/Google.Apis.Assuredworkloads.v1/Google.Apis.Assuredworkloads.v1.cs
```diff
@@ -644,13 +644,88 @@ public ListRequest(Google.Apis.Services.IClientService service, string parent) :
...
                         public virtual string Filter { get; set; }
 
+                        private string _intervalEndTimeRaw;
+
+                        private object _intervalEndTime;
+
                         /// <summary>The end of the time window.</summary>
                         [Google.Apis.Util.RequestParameterAttribute("interval.endTime", Google.Apis.Util.RequestParameterType.Query)]
-                        public virtual object IntervalEndTime { get; set; }
+                        public virtual string IntervalEndTimeRaw
```
amanda-tarafa, line 653: Read first my comment on the doc, but:

I'm not certain what effect will this have on the request "serialization", before it was serializing an object, and now it's serializing a string. I think you need the raw property to control how the value is serialized for the request, but I don't think it needs to be publicly settable. I think users should only set the DTO property, and we set the Raw property based on that with the format we want to serialize the value, and keep the `RequestParameterAttribute` on the raw one so that no more serialization happens.

jskeet: Will make all comments on the doc.

PR #705, opened 2023-11-09
fix: Update the generator to use the new DiscoveryFormat methods

jskeet:
(Only the final commit is new; the rest are from #704.)


amanda-tarafa (approved):
LGTM the last commit

PR #707, opened 2023-11-13
fix: Improvements to date/time handling

### LINKED ISSUE #700: REST: Make date-time and google-datetime format properties call different conversion methods
It looks like these probably have different formats - "date-time" includes a UTC offset, whereas "google-datetime" is always in UTC. Not sure about the precision of either of them - I'd expect "google-datetime" to be nanosecond precision.

jskeet:
This single PR fixes #700 and #701.

It makes #704 and #705 obsolete. (It's basically a rewrite.)

The last commit is the one requiring most concerted review. I'm currently regenerating and rebuilding all packages in google-api-dotnet-client to check for unexpected problems.

### DIFF
--- a/Google.Api.Generator.Rest.Tests/GoldenTestData/Google.Apis.Assuredworkloads.v1/Google.Apis.Assuredworkloads.v1.cs
+++ b/Google.Api.Generator.Rest.Tests/GoldenTestData/Google.Apis.Assuredworkloads.v1/Google.Apis.Assuredworkloads.v1.cs
@@ -1,0 +1,2402 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Generated code. DO NOT EDIT!
+
+namespace Google.Apis.Assuredworkloads.v1
+{
+    /// <summary>The Assuredworkloads Service.</summary>
+    public class AssuredworkloadsService : Google.Apis.Services.BaseClientService
+    {
+        /// <summary>The API version.</summary>
+        public const string Version = "v1";
+
+        /// <summary>The discovery version used to generate this service.</summary>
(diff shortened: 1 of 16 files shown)


Google.Api.Generator.Rest.Tests/GoldenTestData/Google.Apis.Assuredworkloads.v1/Google.Apis.Assuredworkloads.v1.cs
```diff
@@ -644,13 +644,71 @@ public ListRequest(Google.Apis.Services.IClientService service, string parent) :
...
-                        /// <summary>The end of the time window.</summary>
+                        private object _intervalEndTime;
+
+                        /// <summary>
+                        /// String representation of <see cref="IntervalEndTimeDateTimeOffset"/>, formatted for
+                        /// inclusion in the HTTP request.
+                        /// </summary>
                         [Google.Apis.Util.RequestParameterAttribute("interval.endTime", Google.Apis.Util.RequestParameterType.Query)]
-                        public virtual object IntervalEndTime { get; set; }
+                        public virtual string IntervalEndTimeRaw { get; private set; }
```
amanda-tarafa, line 654: Thinking more, do we even need this to be public? I like very much to have this property to apply the attribute to, and we control serialization better. But would users need it at all?

amanda-tarafa (approved):
LGTM, with a question.

Google.Api.Generator.Rest.Tests/GoldenTestData/Google.Apis.Assuredworkloads.v1/Google.Apis.Assuredworkloads.v1.cs
jskeet: Users don't, but we fetch this via reflection, and in some cases I can imagine things breaking if we access internal or private properties via reflection in a different assembly. If we're already doing that elsewhere, I'm happy to include it here too - but I suspect we don't.

jskeet: Righto. Had you looked enough to be really happy for me to merge? This all feels so fiddly I don't want to merge if you'd like any more time at all. (I'm aware you've LGTM'd, I'm just being super-cautious.)

amanda-tarafa: Yep, I think this is fine. I've tried to run over all possible scenarios and I think people that may run into trouble are those that attempt to mix old and new properties. That should be rare, and there's really no reason that I can think so to do so anyway.

jskeet: Great, thanks. Merging.

PR #711, opened 2023-11-15
chore: Update Discovery reference to 1.64.0

jskeet:
At this point, we don't need the Google.Apis 1.64.0-alpha01 reference any more, as we get 1.64.0 transitively.


jskeet:
(This will close #708 and #709.)

amanda-tarafa (approved):

PR #746, opened 2024-02-21
feat: Universe domain support for discovery based libraries.

amanda-tarafa:
This has a dependency on googleapis/google-api-dotnet-client#2675 because generated code uses the methods defined there. This PR's presubmits won't be green until we have merged and released googleapis/google-api-dotnet-client#2675.

### DIFF
--- a/Google.Api.Generator.Rest/Models/PackageModel.cs
+++ b/Google.Api.Generator.Rest/Models/PackageModel.cs
@@ -200,16 +200,14 @@ namespace Google.Api.Generator.Rest.Models
                 var batchUriValue = _discoveryDoc.RootUrl + _discoveryDoc.BatchPath;
                 var batchPathValue = _discoveryDoc.BatchPath;
 
-                var baseUriProperty = Property(Modifier.Public | Modifier.Override, ctx.Type<string>(), "BaseUri")
-                    .WithGetBody(IdentifierName("BaseUriOverride").NullCoalesce(baseUriValue))
+                var baseUriProperty = AutoProperty(Modifier.Public | Modifier.Override, ctx.Type<string>(), "BaseUri", hasSetter: false)
                     .WithXmlDoc(XmlDoc.Summary("Gets the service base URI."));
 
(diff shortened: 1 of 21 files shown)


Google.Api.Generator.Rest/Models/PackageModel.cs
```diff
@@ -218,7 +216,11 @@ private ClassDeclarationSyntax GenerateServiceClass(SourceFileContext ctx)
                 var resourceProperties = Resources.Select(resource => resource.GenerateProperty(ctx)).ToArray();
 
                 var parameterizedCtor = Ctor(Modifier.Public, cls, BaseInitializer(initializerParam))(initializerParam)
-                    .WithBlockBody(resourceProperties.Zip(Resources).Select(pair => pair.First.Assign(New(ctx.Type(pair.Second.Typ))(This))).ToArray())
+                    .WithBlockBody(
+                        resourceProperties.Zip(Resources).Select(pair => pair.First.Assign(New(ctx.Type(pair.Second.Typ))(This)))
+                        .Append(baseUriProperty.Assign(Call("GetEffectiveUri")(IdentifierName("BaseUriOverride"), baseUriValue)))
```
jskeet, line 221: Nit: indent these three lines more? (So they're indented relative to `resourceProperties.Zip`?)

jskeet (approved):
I guess there's a slight memory footprint increase here with more fields rather than calculated properties, but I very much doubt that it'll be significant.

Google.Api.Generator.Rest/Models/PackageModel.cs
amanda-tarafa: Done

amanda-tarafa (commented):
> I guess there's a slight memory footprint increase here with more fields rather than calculated properties, but I very much doubt that it'll be significant.

Yes, I did choose this actively. Calculating the properties now implies replacing "googleapis.com" with the custom domain, if there is one. And that happens for every service call, so I figured having the fields was slightly better. Happy to reconsider. Ultimately I think neither is going to make much of a difference though.

amanda-tarafa:
I've updated to latest GAX and Google.Apis (tests only) latests. I'll merge on green.

PR #772, opened 2024-04-03
feat: Add container access as an extension point to more registration APIs

maw136:
issue: https://github.com/googleapis/gapic-generator-csharp/issues/771

- added necessary generation code
- fixed existing tests that had disabled assertion for generated `ServiceCollectionExtensions.g.cs`
- set invariant culture for running all tests - non english cultured machines failed tests because of different decimal separator


jskeet:
It would really help if all the different bits were in different commits. It's not clear to me at the moment:

- Why you needed to change anything around whether or not extension methods were being tested. What happened without that?
- Whether changing the test context to the invariant culture is the right change, or whether actually that's just masking a bug in the generator.

In general, it's always worth consulting a little before starting on a PR, particularly if you think it will require changes that aren't *immediately* related to the change in question (i.e. "generating more methods in an existing file"). I'll look at the invariance aspect separately now.

jskeet:
(The production code change looks absolutely fine by the way, and thanks for that - it's only the other bits which I'm less sure about.)

jskeet:
Right, between #775 and #776, you should be able to change this PR to *just* changes to the remaining 3 ServiceCollectionExtensions.g.cs test files, and the production code change.

If you need to do *anything* else, please shout and I'll look at what's going on.

jskeet:
Ooh - just catching up on other email, this has a different level of breaking change sensitivity than the Pub/Sub one, because for Pub/Sub the action couldn't be null... here it can (and indeed defaults to null in both overloads, which would cause problems). We probably want to give the new method a different name, although I'm not sure what yet. I'm not sure whether there are existing conventions here, but possibly `AddConfiguredXyzClient`?

maw136:
My thought process is to make minimal amount of changes with maximum outcome so to speak. 
This one is in this order
- I've changed the code generation
- Running unit tests -> problems with float formatting -> added module initialization
- Running unit tests again -> no issues (but there should be because I changed code generation)
- Found an ignore flag to check `ServiceCollectionExtensions` -> enabled the flag to check my change in more (already added) use cases
- Running unit tests -> problems with existing test data -> updated test data to have all green tests

IMHO the PR is self contained and minimal.

maw136:
It's your call as a maintainer. I will adjust, no problem. Could you explain https://github.com/googleapis/gapic-generator-csharp/pull/776 cause I made it work and you deleted the file altogether, I'm just curious.

jskeet:
Thinking about it more... if we generate the new method using the same name, but *without* defaulting (but still *allowing* a null action), I think we might be okay. So we'd generate:

```csharp
public static IServiceCollection AddXyzClient(this IServiceCollection services, sys::Action<sys::IServiceProvider, XyzClientBuilder> action = null)
...

public static IServiceCollection AddXyzClient(this IServiceCollection services, sys::Action<td::XyzClientBuilder> action = null)
```

At this point:

- Current code calling `services.AddXyzClient();` will still resolve to the existing (latter) method
- Current code calling `services.AddXyzClient(someActionOnBuilder);` will still resolve to the existing method
- Current code calling `services.AddXyzClient(null);` will resolve to the *first* method (due to C# overload resolution rules) but it will have the same effect
- Current code calling `services.AddXyzClient(SomeMethodName);` will continue to resolve to the existing method if `SomeMethodName` only has overloads matching `Action<XyzClientBuilder>`; this is a breaking change if `SomeMethodName` also has an overload matching `Action<IServiceProvider, XyzClientBuilder>` at which point *either* it's reported as ambiguous *or* it resolves to the new method calling the other overload... but I'd consider that an extremely unlikely case.

@amanda-tarafa What do you think?

jskeet:
> It's your call as a maintainer. I will adjust, no problem.

My issue with the previous commit was that it was making three sets of changes not just in one PR, but in one commit - which made it harder to reason about them (and to separate out what we wanted to do with them). We try to keep commits as self-contained and single-purpose as is reasonable.

(Sometimes we create a PR with multiple commits which aren't *strictly* related, if that's convenient - other times it's worth having separate PRs. I'm not overly bothered about that side, as it doesn't affect the commit history and it's still easy to review one commit at a time.)

> Could you explain #776 cause I made it work and you deleted the file altogether, I'm just curious.

For each case of "this file exists in the tests, but isn't tested" there are two options:

- We don't actually want to test it, as the test case doesn't affect that file (e.g. publishing settings don't currently affect service the GAPIC API metadata) - in this case, I've deleted the file
- We should be testing it - in this case, I've changed the test to stop ignoring the file

Both options end up with a passing test which does check that the file is as expected, but I wanted to avoid having a lot of files which weren't actually impacted.

In the case of publishing settings, I now think perhaps it's worth reinstating ServiceCollectionExtensions and PackageApiMetadata (but not gapic_metadata) as the client renaming *does* affect those files. I'd suggest we can do that once this PR is all done and dusted though - I'll file a new issue to remind myself to do that.

amanda-tarafa:
@jskeet I've left a note on your original comment, but here also: I updated the code snippet there because the new method was still defaulting the action parameter to null.

> At this point:
> 
> * Current code calling `services.AddXyzClient();` will still resolve to the existing (latter) method
> * Current code calling `services.AddXyzClient(someActionOnBuilder);` will still resolve to the existing method
> * Current code calling `services.AddXyzClient(null);` will resolve to the _first_ method (due to C# overload resolution rules) but it will have the same effect
> * Current code calling `services.AddXyzClient(SomeMethodName);` will continue to resolve to the existing method if `SomeMethodName` only has overloads matching `Action<XyzClientBuilder>`; this is a breaking change if `SomeMethodName` also has an overload matching `Action<IServiceProvider, XyzClientBuilder>` at which point _either_ it's reported as ambiguous _or_ it resolves to the new method calling the other overload... but I'd consider that an extremely unlikely case.

Yep, I agree with all of these. I'd still like to call out the potential breaking change of the last point with the usual note of "we believe it's less disruptive in a minor".

jskeet:
> I'd still like to call out the potential breaking change of the last point with the usual note of "we believe it's less disruptive in a minor".

Definitely.

jskeet:
@maw136: Are you planning to update this PR, or are you happy for us to just start a new one? (That might be simpler.)

maw136:
I will update it according to discussion. Had less time since last post, sorry.

jskeet (approved):
This is great, thanks. I'll squash and merge.

jskeet:
This has now been rolled out, and so will be in the next release of each GAPIC library.

PR #773, opened 2024-04-03
bugfix: Fix xunit output - xunit doesn't capture Console output

maw136:


jskeet:
Closing having merged #774. There's no need for the extra complexity just to get additional output we're not using.

maw136:
Actually the output was helpful to get the full generated file. I used to fix some failing tests.

jskeet:
> Actually the output was helpful to get the full generated file. I used to fix some failing tests.

The full generated files are created under the temp directory, in a complete directory structure within a directory called "GeneratorTests-" followed by a timestamp.

maw136:
Ok, good to know.

PR #779, opened 2024-04-04
Lro method settings

### LINKED ISSUE #557: Make LRO polling generation configurable
We currently use a hard-coded set of LRO polling options.

There *is* configuration for LRO polling, but it's in GAPIC YAML files which we don't currently use for anything else. Given the lack of customer feedback on this, we can probably leave this for now - but we should try to rationalize where our configuration comes from, and then include LRO polling in that.

jskeet:
Fixes #557.

I'm going to run this locally against all APIs to see whether we get any changes - I don't *expect* there to be any settings configured in MethodSettings yet. (They're all in _gapic.yaml files.)

We should then discuss what we want to do.

### DIFF
--- a/Google.Api.Generator.Tests/ProtoTests/Lro/Lro.proto
+++ b/Google.Api.Generator.Tests/ProtoTests/Lro/Lro.proto
@@ -28,8 +28,8 @@ service Lro {
     option (google.api.method_signature) = "name";
   }
 
-  // Test an LRO RPC with customized default polling settings.
-  rpc CustomDefaultPollingMethod(Request) returns(google.longrunning.Operation) {
+  // Test an LRO RPC with polling settings specified in the service config.
+  rpc ServiceConfigPollingMethod(Request) returns(google.longrunning.Operation) {
     option (google.longrunning.operation_info) = {
       response_type: "LroResponse"
       metadata_type: "LroMetadata"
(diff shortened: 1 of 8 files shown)


Google.Api.Generator.Tests/ProtoTests/Lro/Lro.proto
```diff
@@ -27,6 +27,15 @@ service Lro {
     };
     option (google.api.method_signature) = "name";
   }
+
+  // Test an LRO RPC with customized default polling settings.
+  rpc CustomDefaultPollingMethod(Request) returns(google.longrunning.Operation) {
```
amanda-tarafa, line 32: I know this is just a test method, but the combination of custom and default is always very confusing to me. Can you use default override instead?

amanda-tarafa (commented):
Yep, this looks good.

Google.Api.Generator.Tests/ProtoTests/Lro/Lro.proto
jskeet: Or maybe PublishingSettingsPollingMethod? ServiceConfigPollingMethod? Happy to bike shed this :)

amanda-tarafa (approved):

jskeet:
Generation has completed - it would lead to no changes in google-cloud-dotnet.

amanda-tarafa (dismissed):

jskeet:
Will have a look tomorrow. I'm not actually in a hurry to get this in, but I wanted it in a better shape :)

Issue #652, opened 2023-08-03
Fail generation if any protos have missing comments


jskeet:
Currently, if any protobuf element is missing a comment, we generate code but the result doesn't compile. The missing comments could be on messages, RPCs, services, fields, enums, enum values - anything, basically.

It would be helpful for the generator to fail for this:

- We would highlight the problem to API producers earlier
- We can provide an error message with the complete list of elements missing comments, rather than having to look through protos after finding C# compilation errors

PR #792, opened 2024-04-23
feat: Log warnings and errors on missing proto comments

jskeet:
Towards #652.

There are three potential levels of enforcement:

- Level 0 (this commit): log warnings and errors, but don't actually make the generator fail
- Level 1: fail if a service or RPC has no comment, as that will definitely break the generated GAPIC code
- Level 2: fail if *any* comment is missing

Level 2 would be a nice ideal, but level 1 is more practical. This commit doesn't go that far, as we suspect there will be a number of APIs where that would fail. The first step should be to generate as many APIs as we can (not just those in google-cloud-dotnet) and find out how many *would* fail. Ideally, we can then get those fixed, and turn on level 1 enforcement.

### DIFF
--- a/Google.Api.Generator/CodeGenerator.cs
+++ b/Google.Api.Generator/CodeGenerator.cs
@@ -189,6 +189,7 @@ namespace Google.Api.Generator
             }
 
             ValidateTransports(transports, allServiceDetails);
+            ValidateComments(descriptors.Where(x => filesToGenerateSet.Contains(x.Name)));
 
             Logging.LogTrace("Generation complete with {count} files", resultFiles.Count);
             return resultFiles;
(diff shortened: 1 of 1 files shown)


Google.Api.Generator/CodeGenerator.cs
```diff
@@ -236,6 +237,103 @@ private static void ValidateTransports(ApiTransports transports, List<ServiceDet
             }
         }
 
+        /// <summary>
+        /// Validate that all proto elements have comments (except the file descriptor).
+        /// We only log warnings for most missing elements; for services and methods, we log warnings
+        /// and then fail with an exception at the end, as the generated code won't compile.
```
amanda-tarafa, line 243: This is not the case as the code is at the moment. Maybe say something like 
```suggestion
        /// and in the future we may fail with an exception at the end, as the generated code won't compile.
```

amanda-tarafa (approved):

PR #794, opened 2024-04-30
feat: Validate the choice of paginated resource field

jskeet:
This check will potentially be removed later, if we get explicit configuration.

It's almost certainly "correct by construction" from all the conditions where we call the constructor, but it's handy to validate in a single location.

Fixes b/337763481.


jskeet:
(I've run this locally against google-cloud-dotnet, and everything is fine.)

amanda-tarafa (approved):

PR #801, opened 2024-05-15
feat: Generate properties for nested resource reference fields

jskeet:
Currently generating google-cloud-dotnet so I can post a diff there (although it'll also include all the reordering).

Fixes b/339353198.

### DIFF
--- a/Google.Api.Generator/Generation/ResourceNamesGenerator.cs
+++ b/Google.Api.Generator/Generation/ResourceNamesGenerator.cs
@@ -524,23 +524,41 @@ namespace Google.Api.Generator.Generation
         {
             // Note: Whether a field is required or optional is purposefully ignored in this partial class.
             // The optionalness of a field is only relevant within a method signature (flattening).
-            foreach (var msg in _fileDesc.MessageTypes)
+            foreach (var topLevelMessage in _fileDesc.MessageTypes)
             {
-                var resources = msg.Fields.InFieldNumberOrder()
-                    .Select(fieldDesc => (fieldDesc, resDetails: _catalog.GetResourceDetailsByField(fieldDesc)))
-                    .Where(x => x.resDetails != null)
+                // First find all resource-referencing fields for all messages recursively.
+                // We then pick only the relevant ones to generate on a per-message basis.
+                var allResources = topLevelMessage.SelfAndNestedMessagesRecursively()
+                    .SelectMany(m => m.Fields.InFieldNumberOrder())
+                    .Select(_catalog.GetResourceDetailsByField)
+                    .Where(x => x is not null)
+                    .SelectMany(r => r)
                     .ToList();
-                if (resources.Any())
+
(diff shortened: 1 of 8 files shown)


Google.Api.Generator/Generation/ResourceNamesGenerator.cs
```diff
@@ -524,23 +524,41 @@ private IEnumerable<ClassDeclarationSyntax> ProtoMessagePartials()
...
+                    .Select(r => r.Descriptor.ContainingType)
+                    .SelectMany(m => m.SelfAndAncestors())
+                    .ToHashSet();
+
+                if (allResources.Any())
+                {
+                    yield return GenerateClassWithResourceReferences(topLevelMessage);
+                }
+
+                ClassDeclarationSyntax GenerateClassWithResourceReferences(MessageDescriptor msg)
```
amanda-tarafa, line 548: Is this capturing anything from within the foreach scope? If not, maybe declare it outside the foreach, even if for dropping the extra indent.

Google.Api.Generator/ProtoUtils/ProtoExtensions.cs
```diff
@@ -78,6 +78,32 @@ internal static string RemoveEnumPrefix(string enumName, string valueName)
...
         }
 
+        /// <summary>
+        /// Returns a sequence of messages covering <paramref name="descriptor"/> and all nested messages within it,
+        /// recursively.
+        /// </summary>
+        internal static IEnumerable<MessageDescriptor> SelfAndNestedMessagesRecursively(this MessageDescriptor descriptor)
+        {
+            yield return descriptor;
+            foreach (var nestedType in descriptor.NestedTypes.SelectMany(SelfAndNestedMessagesRecursively))
```
amanda-tarafa, line 88: Can NestedTypes have enums or just messages. It doesn't matter a lot because the enums won't have resource properties, but it'd be at odds with this method name.

amanda-tarafa (approved):
LGTM, I've made a couple of comments but I'm happy to merge as is.

Google.Api.Generator/Generation/ResourceNamesGenerator.cs
jskeet: Will give that a try.

Google.Api.Generator/ProtoUtils/ProtoExtensions.cs
jskeet: NestedTypes is only messages. (There's a separate NestedEnums property.)

Google.Api.Generator/Generation/ResourceNamesGenerator.cs
jskeet: It captures `allResources` and `messagesContainingResources` which are both declared in the loop. (We could potentially make them file-descriptor wide, but I don't think it's worth it.) Merging as-is.

PR #816, opened 2024-06-27
feat: Fail generation when a service or method is missing a comment

### LINKED ISSUE #652: Fail generation if any protos have missing comments

jskeet:
Fixes #652.

### DIFF
--- a/Google.Api.Generator.Tests/ProtoTest.proto
+++ b/Google.Api.Generator.Tests/ProtoTest.proto
@@ -2,7 +2,9 @@
 
 package testing;
 
+// Test service
 service Test {
+  // Test method
   rpc Method(Msg) returns(Msg);
 }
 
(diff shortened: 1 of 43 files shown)


jskeet:
Note: I've tested this manually by running the generator against a deliberately-missing-a-comment API. It's relatively tricky for us to add one-off failure tests... something to think about in the future, maybe.

amanda-tarafa (approved):

PR #820, opened 2024-07-24
fix: Conditionally specify page token and page size in pagination

jskeet:
I'm going to run this locally against everything in google-cloud-dotnet, and build the result.


jskeet:
All regenerated and tested fine. The only other changes since our last release are dependency updates, so if you're happy with this @amanda-tarafa, we should be good to release and adopt immediately.

amanda-tarafa (approved):
Yep, lgtm!

PR #844, opened 2024-10-03
Initial Commit

sakib-dac:
Allow ConsoleApp to accept two arguments only (<Discovery Doc JSON file> <output directory> )
Generate only the DataModels.
Create Separate cs Files for each of the Model Classes.


sakib-dac:
wrong PR.

PR #846, opened 2024-10-18
feat: Support editions (up to edition 2023)

jskeet:
There are no generator changes beyond declaring support in the response. I don't believe we use field presence anywhere in the generator (having deliberately looked for that). (It's possible that we *should* when generating method signatures for RPCs, but that's a different matter.)

I expect we'll want to integrate Showcase tests for editions as well.


amanda-tarafa (approved):
That was short :)

PR #860, opened 2025-01-16
feat: Generate enum strings for Compute

jskeet:
This moves code from google-cloud-dotnet to the generator, removing the post-processing.

We can probably adopt this as-is for the moment (given that this is only for DIREGAPIC, and only Compute uses DIREGAPIC) but we should really have configuration within the service config publish settings to turn this feature on (and potentially configure the filename; that'll be easier to tell if we end up with more examples).

### DIFF
--- a/Google.Api.Generator/Generation/EnumStringGenerator.cs
+++ b/Google.Api.Generator/Generation/EnumStringGenerator.cs
@@ -1,0 +1,148 @@
+﻿// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using Google.Api.Generator.ProtoUtils;
+using Google.Api.Generator.Utils;
+using Google.Api.Generator.Utils.Roslyn;
+using Google.Protobuf.Reflection;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using static Google.Api.Generator.Utils.Roslyn.Modifier;
+using static Google.Api.Generator.Utils.Roslyn.RoslynBuilder;
+using static Microsoft.CodeAnalysis.CSharp.SyntaxFactory;
+
+namespace Google.Api.Generator.Generation;
+
+/// <summary>
+/// Generates a static class (with nested static classes) to provide
+/// string constants for the wire representations of enum values. This is only used
+/// for DIREGAPIC APIs (currently just Compute).
+/// </summary>
+internal class EnumStringGenerator
+{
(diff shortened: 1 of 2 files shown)


jskeet:
(Note that we could remove *one* piece of config by generating one type per FileDescriptor, and using the name of the FileDescriptor as a prefix with a hard-coded suffix. Compute only has `compute.proto`. Let me know if you want me to implement that.

Testing: this is basically tested by running it against the Compute API and checking that the result is "identical" (in semantic diff terms) to the current code.

Google.Api.Generator/Generation/EnumStringGenerator.cs
```diff
@@ -0,0 +1,152 @@
...
+/// </summary>
+internal class EnumStringGenerator
+{
+    // TODO: Determine a policy or config for the class, file and API description.
+    internal const string RootClassName = "ComputeEnumConstants";
+    private const string ApiDescription = "Compute API";
+
+    public static CompilationUnitSyntax MaybeGenerate(SourceFileContext ctx, string ns, IEnumerable<FileDescriptor> packageFileDescriptors)
+    {
+        // TODO: Add an option in the service config for this... although
```
amanda-tarafa, line 42: nit: seems incomplete... although I'll probably be happy with what comes after the although :)

Google.Api.Generator/Generation/EnumStringGenerator.cs
```diff
@@ -0,0 +1,152 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using static Google.Api.Generator.Utils.Roslyn.Modifier;
+using static Google.Api.Generator.Utils.Roslyn.RoslynBuilder;
+using static Microsoft.CodeAnalysis.CSharp.SyntaxFactory;
+
+namespace Google.Api.Generator.Generation;
+
+/// <summary>
+/// Generates (to stdout) a static class (with nested static classes) to provide
```
amanda-tarafa, line 30:
```suggestion
/// Generates a static class (with nested static classes) to provide
```

Google.Api.Generator/Generation/EnumStringGenerator.cs
```diff
@@ -0,0 +1,152 @@
...
+/// string constants for the wire representations of enum values. This is only used
+/// for DIREGAPIC APIs.
```
amanda-tarafa, line 32:
```suggestion
/// for DIREGAPIC APIs which is just the Compute API.
```

Google.Api.Generator/Generation/EnumStringGenerator.cs
```diff
@@ -0,0 +1,152 @@
...
+
+        public bool ShouldGenerate => Enums.Count > 0 || NestedContainers.Any(nc => nc.ShouldGenerate);
+
+        public EnumContainer(MessageDescriptor descriptor, IEnumerable<EnumDescriptor> enums, IEnumerable<MessageDescriptor> messages)
+        {
+            Descriptor = descriptor;
+            if (Descriptor is not null)
+            {
+                Utils.Logging.LogWarning("Typ of {name} is {fullname}", Descriptor.FullName, ProtoTyp.Of(Descriptor).FullName);
+                //throw new Exception($"Typ of {Descriptor.FullName} is {ProtoTyp.Of(Descriptor).FullName}");
```
amanda-tarafa, line 81: nit: leftover

amanda-tarafa (approved):
A few nits, but LGTM.

I'm totally happy with this as is, we can figure things out if we ever need to.

amanda-tarafa (approved):

PR #875, opened 2025-06-04
feat: Loosen constraints on pagination heuristics

jskeet:
This should unblock recent Compute changes. It may well also be looser than we'd ideally like it to be in general, but shouldn't any existing APIs. Anything that *already* generates successfully should continue to generate with the same result, but there may be APIs that we *don't* want to generate successfully that will now do so.


jskeet:
Marking as draft as I'm checking what it does to any other APIs at the moment.

PR #878, opened 2025-06-09
chore: Initial setup for releasing from Librarian

jskeet:
Note that this resets the Version property in project files to the last-released version, so that we can run the release process to actually release version 1.5.0.

This requires https://github.com/googleapis/google-cloud-dotnet/pull/14860 to be merged and set as the latest container image.


amanda-tarafa (approved):
At a glance this looks good.

PR #883, opened 2025-07-08
fix: avoid bazel rules failing due to unexpected dotnet output

jskeet:
The first run of "dotnet" in any given location includes output of:

> An issue was encountered verifying workloads. For more information,
> run "dotnet workload update"

... which is unhelpful when stdout is being used by protoc.

This is a hacky fix which just runs "dotnet workload update" first to get this output out of the way. (It doesn't actually fix the workload verification, but it avoids the subsequent output.)


jskeet:
(This isn't in my fork as it's easier to test Bazel changes when they're just in branches of the main repo.)

jskeet:
I now have an alternative to try...

PR #886, opened 2025-07-08
test: add presubmit test for testing bazel

chingor13:
This presubmit should help us be confident that it will run against googleapis/googleapis when we try to update the generator version.

### DIFF
--- a/cloudbuild-test.yaml
+++ b/cloudbuild-test.yaml
@@ -1,0 +1,28 @@
+# Copyright 2025 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+timeout: 7200s # 2 hours
+
+steps:
+
+- id: googleapis-test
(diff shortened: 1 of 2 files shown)


cloudbuild-test.yaml
```diff
@@ -0,0 +1,28 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+timeout: 7200s # 2 hours
+
+steps:
+
+- id: googleapis-test
+  name: gcr.io/gapic-images/googleapis:prod
+  entrypoint: "/workspace/tools/bazel-test.sh"
```
amanda-tarafa, line 21: I think?
```suggestion
  entrypoint: "/tools/bazel-test.sh"
```

cloudbuild-test.yaml
```diff
@@ -0,0 +1,28 @@
+# Copyright 2025 Google LLC
```
amanda-tarafa, line 1: Does this need to be on the repo root? I don't mind that much but if it can be in a folder similar to the actions folder that'd be good, like `.cloudbuild`

tools/bazel-test.sh
```diff
@@ -0,0 +1,57 @@
...
+
+set -eo pipefail
+
+apt-get update -y
+apt-get install -y jq
+
+GAPIC_GENERATOR_COMMIT=$1
+if [[ -z "${GAPIC_GENERATOR_COMMIT}" ]]
+then
+  echo "GAPIC_GENERATOR_COMMIT not set, trying default"
```
amanda-tarafa, line 24:
```suggestion
  echo "GAPIC_GENERATOR_COMMIT not set"
```

amanda-tarafa (approved):
A couple of comments, but likely because I don't have enough context.

amanda-tarafa:
I'll close this one, it's been sitting here for long. We can recreate it if we need to.

PR #990, opened 2026-08-05
fix: suppress warnings on enumeration of deprecated fields

robertvoinescu-work:
b/543020713


amanda-tarafa (changes requested):
Please add tests for this.

PR #997, opened 2026-08-25
ci: Add integration test runner scripts for GAPIC Showcase

robertvoinescu-work:
b/522438834

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,7 @@ bazel-*
 tmp
 launchSettings.json
 .cr/
+/gapic-showcase
+/gapic-showcase.exe
+/showcase.pid
+/showcase.log
--- a/runintegrationtests.sh
+++ b/runintegrationtests.sh
@@ -1,0 +1,44 @@
+#!/bin/bash
+
+# This script:
+# - Builds the Generator solution in Release mode
+# - Starts GAPIC Showcase
+# - Runs integration tests for the Generator projects in Release mode
+# - Tears down GAPIC Showcase on exit
+
+set -e
+
+cd "$(dirname "$0")"
+
+export DOTNET_NOLOGO=true
+
+CONFIG=Release
+DOTNET_BUILD_ARGS="-c $CONFIG"
+DOTNET_TEST_ARGS="--no-build $DOTNET_BUILD_ARGS"
+
+echo "Building the solution in $CONFIG mode"
+dotnet build $DOTNET_BUILD_ARGS Generator.sln
+
+cleanup() {
+  echo "Tearing down GAPIC Showcase"
+  if [ -f showcase.pid ]; then
+    kill "$(cat showcase.pid)" 2>/dev/null || true
+    rm -f showcase.pid
+  fi
+  rm -f gapic-showcase gapic-showcase.exe
+}
+trap cleanup EXIT
+
+echo "Setup GAPIC Showcase for standard integration tests"
+./startshowcase.sh --port :7469
+
+export SHOWCASE_ENDPOINT=http://localhost:7469
+dotnet test $DOTNET_TEST_ARGS Google.Api.Generator.IntegrationTests
+
+echo "Setup GAPIC Showcase with TLS for PQC integration tests"
+./startshowcase.sh --port :7469 --tls
+
+export SHOWCASE_ENDPOINT=https://localhost:7469
+dotnet test $DOTNET_TEST_ARGS Google.Api.Generator.IntegrationTests --filter "FullyQualifiedName~Pqc"
+
+echo "Integration testing completed"
--- a/startshowcase.sh
+++ b/startshowcase.sh
@@ -1,0 +1,52 @@
+#!/bin/bash
+set -eo pipefail
+
+cd "$(dirname "$0")"
+
+if [[ "$OSTYPE" == "linux-gnu"* ]]; then OS="linux"; elif [[ "$OSTYPE" == "darwin"* ]]; then OS="darwin"; else OS="windows"; fi
+
+ARCH="amd64"
+if [[ "$OS" == "darwin" && $(uname -m) == "arm64" ]]; then
+  ARCH="arm64"
+fi
+
+# Clean up any previously running instance
+if [ -f showcase.pid ]; then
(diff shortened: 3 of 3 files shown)


startshowcase.sh
```diff
@@ -0,0 +1,46 @@
...
+  echo "Downloading gapic-showcase-${SHOWCASE_VERSION}-${OS}-${ARCH}..."
+  curl -sSL -f https://github.com/googleapis/gapic-showcase/releases/download/v${SHOWCASE_VERSION}/gapic-showcase-${SHOWCASE_VERSION}-${OS}-${ARCH}.tar.gz | tar -zx
+fi
+
+ARGS="${@:---port :7469}"
+if [[ "$OS" == "windows" ]]; then
+  ./gapic-showcase.exe run $ARGS > showcase.log 2>&1 &
+else
+  setsid ./gapic-showcase run $ARGS > showcase.log 2>&1 &
+fi
```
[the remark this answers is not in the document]
robertvoinescu-work: Good catch, fixed.

runintegrationtests.sh
```diff
@@ -0,0 +1,44 @@
...
+dotnet build $DOTNET_BUILD_ARGS Generator.sln
+
+cleanup() {
+  echo "Tearing down GAPIC Showcase"
+  if [ -f showcase.pid ]; then
+    kill $(cat showcase.pid) 2>/dev/null || true
+    rm -f showcase.pid
+  fi
+  rm -f gapic-showcase gapic-showcase.exe
+}
```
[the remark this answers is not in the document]
robertvoinescu-work: Keeping as is - this way we can avoid having a stale gapic generator binary when running locally. It should not affect the CI runs.

startshowcase.sh
```diff
@@ -0,0 +1,46 @@
...
+if [ -f showcase.pid ]; then
+  kill $(cat showcase.pid) 2>/dev/null || true
+  rm -f showcase.pid
+fi
+pkill -f "gapic-showcase run" 2>/dev/null || true
+sleep 1
+
+if [ ! -f gapic-showcase ] && [ ! -f gapic-showcase.exe ]; then
+  echo "Resolving the latest GAPIC Showcase version for $OS-$ARCH..."
+  SHOWCASE_VERSION=$(curl -s https://api.github.com/repos/googleapis/gapic-showcase/releases/latest | sed -n 's/.*"tag_name": "v\([^"]*\)".*/\1/p')
```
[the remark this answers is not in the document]
robertvoinescu-work: Done.

startshowcase.sh
```diff
@@ -0,0 +1,51 @@
...
+fi
+pkill -f "gapic-showcase run" 2>/dev/null || true
+sleep 1
+
+if [ ! -f gapic-showcase ] && [ ! -f gapic-showcase.exe ]; then
+  echo "Resolving the latest GAPIC Showcase version for $OS-$ARCH..."
+  SHOWCASE_VERSION=$(curl -s https://api.github.com/repos/googleapis/gapic-showcase/releases/latest | sed -n 's/.*"tag_name": "v\([^"]*\)".*/\1/p' || true)
+  if [[ -z "$SHOWCASE_VERSION" ]]; then
+    # Fallback to known stable version if GitHub API rate limit is exceeded
+    SHOWCASE_VERSION="0.43.0"
```
efevans, line 26: Could we output a warning here that we are falling back to a potentially old version, I feel that could speedup troubleshooting in the future especially as Github has had some reliability issues as of late

runintegrationtests.sh
```diff
@@ -0,0 +1,44 @@
...
+./startshowcase.sh --port :7469
+
+export SHOWCASE_ENDPOINT=http://localhost:7469
+dotnet test $DOTNET_TEST_ARGS Google.Api.Generator.IntegrationTests
+
+echo "Setup GAPIC Showcase with TLS for PQC integration tests"
+./startshowcase.sh --port :7469 --tls
+
+export SHOWCASE_ENDPOINT=https://localhost:7469
+dotnet test $DOTNET_TEST_ARGS Google.Api.Generator.IntegrationTests --filter "FullyQualifiedName~Pqc"
```
efevans, line 42: Is this intended to be the only test being run?

efevans (changes requested):

startshowcase.sh
robertvoinescu-work: Good points, github reliability has not been too great. Done.

runintegrationtests.sh
robertvoinescu-work: Yes, all other integration tests are run run above without TLS. PQC works over TLS because it uses a custom validation bypass (DangerousAcceptAnyServerCertificateValidator) to accept Showcase's self-signed cert. We can't do this for the other tests because they test the actual generated client builders, which don't expose a hook to bypass TLS certificate validation.

efevans (approved):

runintegrationtests.sh
amanda-tarafa, line 42: Everything should run now under TLS, that's closer to real life.

```suggestion
echo "Setup GAPIC Showcase"
./startshowcase.sh --port :7469 --tls

export SHOWCASE_ENDPOINT=https://localhost:7469
dotnet test $DOTNET_TEST_ARGS Google.Api.Generator.IntegrationTests
```

.gitignore
```diff
@@ -11,3 +11,7 @@ bazel-*
 tmp
 launchSettings.json
 .cr/
+/gapic-showcase
+/gapic-showcase.exe
+/showcase.pid
+/showcase.log
```
amanda-tarafa, line 17: Can't we clean up after ourselves instead?

startshowcase.sh
```diff
@@ -0,0 +1,52 @@
...
+
+# Clean up any previously running instance
+if [ -f showcase.pid ]; then
+  kill "$(cat showcase.pid)" 2>/dev/null || true
+  rm -f showcase.pid
+fi
+pkill -f "gapic-showcase run" 2>/dev/null || true
+sleep 1
+
+if [ ! -f gapic-showcase ] && [ ! -f gapic-showcase.exe ]; then
```
amanda-tarafa, line 21: We probably want to always run with the latest version?

PR #1005, opened 2026-09-17
test: Enable TLS for all Integration Tests and run PQC against generated Client

robertvoinescu-work:
b/484351289

## Validation
- [x] Linux Tests Run is Green
image

- [x] Linux PQC Tests Pass
image

---

- [x] Windows Test Run is Green
image

- [x] Windows PQC Tests Pass
image


Google.Api.Generator.IntegrationTests/ShowcaseTestBase.cs
```diff
@@ -40,8 +41,26 @@ protected ShowcaseTestBase()
             // (We could add an attribute instead, but just going by convention is simpler.)
             _adapter = GetType().Name switch
             {
+                // RestGrpcAdapter does not forward GrpcChannelOptions.HttpHandler to its internal HttpClient,
+                // so REST tests require the SSL_CERT_FILE environment variable (set in runintegrationtests.sh)
+                // to trust Showcase's self-signed certificate.
                 string name when name.EndsWith("RestTest") => RestGrpcAdapter.Default,
```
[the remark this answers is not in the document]
robertvoinescu-work: Not applicable anymore. To get around the windows limitation I've uniformly modified both handlers to accept any certificate without checking it is trusted (SSL_CERT_FILE or within the cert store). This allows us to avoid installing the cert to the root certificate store on windows.

Google.Api.Generator.IntegrationTests/ShowcaseTestBase.cs
```diff
@@ -55,12 +57,46 @@ protected TClient CreateClient()
...
+        private static void AcceptAnyCertificate(ChannelBase channel)
+        {
+            // No functionality is exposed to modify the internal HTTP handler to bypass certificate validation.
+            // Using reflection allows us to avoid installing the Showcase certificate into the OS root store;
+            // TLS encryption still occurs, we just skip verifying the certificate's Certificate Authority.
+            var invoker = (channel.GetType().GetField("_httpClient", BindingFlags.Instance | BindingFlags.NonPublic)
+                ?? channel.GetType().GetField("<HttpInvoker>k__BackingField", BindingFlags.Instance | BindingFlags.NonPublic))
+                ?.GetValue(channel) as HttpMessageInvoker;
+
+            var handler = typeof(HttpMessageInvoker).GetField("_handler", BindingFlags.Instance | BindingFlags.NonPublic)?.GetValue(invoker);
```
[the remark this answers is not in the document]
robertvoinescu-work: I've added the null checks.

Google.Api.Generator.IntegrationTests/PqcIntegrationTest.cs
```diff
@@ -12,104 +12,43 @@
...
-            var invoker = channel.CreateCallInvoker();
-            using var call = invoker.AsyncUnaryCall(method, null, new CallOptions(), Array.Empty<byte>());
-            await call.ResponseAsync;
             var allHeaders = (await call.ResponseHeadersAsync).Concat(call.GetTrailers());
 
-            // Retrieve the client's advertised cipher list (supported) and the server's chosen cipher (negotiated).
-            // We expect exactly one entry for each header, throwing otherwise.
             var clientSupportedGroupsEntry = Assert.Single(allHeaders, e => e.Key.Equals("x-showcase-tls-client-supported-groups", StringComparison.OrdinalIgnoreCase));
+            Skip.If(!clientSupportedGroupsEntry.Value.Contains("MLKEM", StringComparison.OrdinalIgnoreCase),
+                $"Skipping: We rely on the host OS for protocol support. Unable to verify that an allowed PQC protocol (ML-KEM) is used when present because the host does not advertise it. Advertised: {clientSupportedGroupsEntry.Value}");
```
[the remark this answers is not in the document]
robertvoinescu-work: Resolved.

Google.Api.Generator.IntegrationTests/ShowcaseTestBase.cs
```diff
@@ -55,12 +57,52 @@ protected TClient CreateClient()
...
+            {
+                case SocketsHttpHandler socketsHandler: // Used by GrpcChannel
+                    socketsHandler.SslOptions.RemoteCertificateValidationCallback = (_, _, _, _) => true;
+                    break;
+                case HttpClientHandler httpClientHandler: // Used by RestChannel
+                    httpClientHandler.ServerCertificateCustomValidationCallback = HttpClientHandler.DangerousAcceptAnyServerCertificateValidator;
+                    break;
+                default:
+                    throw new InvalidOperationException($"Unsupported HTTP handler type: {handler?.GetType().FullName}");
+            }
```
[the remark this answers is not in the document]
robertvoinescu-work: This is a clearer message - done.

Google.Api.Generator.IntegrationTests/PqcIntegrationTest.cs
```diff
@@ -12,104 +12,43 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-using Grpc.Core;
-using Grpc.Net.Client;
+using Google.Showcase.V1Beta1;
```
efevans: Is this a new dependency? I'm getting `The type or namespace name 'Showcase' does not exist in the namespace 'Google' (are you missing an assembly reference?)` here

Google.Api.Generator.IntegrationTests/PqcIntegrationTest.cs
```diff
@@ -12,104 +12,43 @@
...
+                $"Skipping: We rely on the host OS for protocol support. Unable to verify that an allowed PQC protocol (ML-KEM) is used when present because the host does not advertise it. Advertised: {clientSupportedGroupsEntry.Value}");
+
             var negotiatedGroupEntry = Assert.Single(allHeaders, e => e.Key.Equals("x-showcase-tls-group", StringComparison.OrdinalIgnoreCase));
-            
+
             // 'MLKEM' substring confirms post-quantum encryption despite naming variations.
             // Note: If new post-quantum algorithms are standardized in the future, these assertions may require updating.
             // See https://en.wikipedia.org/wiki/Post-Quantum_Cryptography_Standardization
             Assert.Contains("MLKEM", clientSupportedGroupsEntry.Value, StringComparison.OrdinalIgnoreCase);
             Assert.Contains("MLKEM", negotiatedGroupEntry.Value, StringComparison.OrdinalIgnoreCase);
```
efevans, line 46: Do we care that the showcase server picked an `MKLEM` algorithm, as long as the client is sending it? The PR generally looks good but I'm wondering what we're getting out of this test since it's being skipped in the case where the client doesn't offer `MKLEM` algorithm, after that point aren't we just testing the showcase server behavior?

Google.Api.Generator.IntegrationTests/PqcIntegrationTest.cs
```diff
@@ -12,104 +12,43 @@
...
+            Skip.If(clientSupportedGroupsEntry.Value.IndexOf("MLKEM", StringComparison.OrdinalIgnoreCase) < 0,
+                $"Skipping: We rely on the host OS for protocol support. Unable to verify that an allowed PQC protocol (ML-KEM) is used when present because the host does not advertise it. Advertised: {clientSupportedGroupsEntry.Value}");
+
             var negotiatedGroupEntry = Assert.Single(allHeaders, e => e.Key.Equals("x-showcase-tls-group", StringComparison.OrdinalIgnoreCase));
-            
+
             // 'MLKEM' substring confirms post-quantum encryption despite naming variations.
             // Note: If new post-quantum algorithms are standardized in the future, these assertions may require updating.
             // See https://en.wikipedia.org/wiki/Post-Quantum_Cryptography_Standardization
             Assert.Contains("MLKEM", clientSupportedGroupsEntry.Value, StringComparison.OrdinalIgnoreCase);
```
efevans: NIT: remove this if we're skipping on it above

Google.Api.Generator.IntegrationTests/PqcIntegrationTest.cs
robertvoinescu-work: I think we aren't just testing showcase behavior in that case - but we are dependent on it. We are testing the full integration of PQC protocol within our client code ensuring we are able to properly encrypt and  we default to PQC when multiple protocols are available (e.g. potentially an error could cause us to fallback to classical curves).

Google.Api.Generator.IntegrationTests/PqcIntegrationTest.cs
robertvoinescu-work: Good point removed.

Google.Api.Generator.IntegrationTests/PqcIntegrationTest.cs
efevans: Ah so when you call out "default to PQC", that means the supported group was ordered such that an `MKLEM` algorithm was not just present but also top priority?

robertvoinescu-work: Correct, it's what was chosen and it was successfully utilized on both ends.

efevans (approved):
LGTM

Google.Api.Generator.IntegrationTests/ShowcaseTestBase.cs
```diff
@@ -55,12 +57,57 @@ protected TClient CreateClient()
...
+
+        private static void AcceptAnyCertificate(ChannelBase channel)
+        {
+            if (channel is null)
+            {
+                return;
+            }
+            // No functionality is exposed to modify the internal HTTP handler to bypass certificate validation.
+            // Using reflection allows us to avoid installing the Showcase certificate into the OS root store;
+            // TLS encryption still occurs, we just skip verifying the certificate's Certificate Authority.
```
amanda-tarafa: Yikes, I dislike this a lot. How hard is it to install the certificate we get from Showcase?

Google.Api.Generator.IntegrationTests/PqcIntegrationTest.cs
```diff
@@ -12,104 +12,42 @@
...
-            using var call = invoker.AsyncUnaryCall(method, null, new CallOptions(), Array.Empty<byte>());
-            await call.ResponseAsync;
             var allHeaders = (await call.ResponseHeadersAsync).Concat(call.GetTrailers());
 
-            // Retrieve the client's advertised cipher list (supported) and the server's chosen cipher (negotiated).
-            // We expect exactly one entry for each header, throwing otherwise.
             var clientSupportedGroupsEntry = Assert.Single(allHeaders, e => e.Key.Equals("x-showcase-tls-client-supported-groups", StringComparison.OrdinalIgnoreCase));
+            Skip.If(clientSupportedGroupsEntry.Value.IndexOf("MLKEM", StringComparison.OrdinalIgnoreCase) < 0,
+                $"Skipping: We rely on the host OS for protocol support. Unable to verify that an allowed PQC protocol (ML-KEM) is used when present because the host does not advertise it. Advertised: {clientSupportedGroupsEntry.Value}");
+
```
amanda-tarafa: Have we run these tests in any environment that we know supports PQC?

amanda-tarafa (commented):
This looks good, but I have two important questions.

Google.Api.Generator.IntegrationTests/PqcIntegrationTest.cs
robertvoinescu-work: Yes, see images attached above for validation where PQC is supported on the platform. 

I've also now added github actions and they are running (not skipped) there also and passing for both linux and windows.

Google.Api.Generator.IntegrationTests/ShowcaseTestBase.cs
robertvoinescu-work: Thanks for pushing back. I spent a bit more time digging and we can actually avoid installing to the root certificate store avoiding the need for admin permissions. See github actions.

Google.Api.Generator.IntegrationTests/ResumableUploadRestTest.cs
```diff
@@ -26,7 +26,7 @@ namespace Google.Api.Generator.IntegrationTests
 {
     public abstract class ResumableUploadTestBase : ShowcaseTestBase<ResumableUploadServiceClient, ResumableUploadServiceClientBuilder>
     {
-        [SkippableFact]
+        [SkippableFact(Skip = "b/556694369")]
```
amanda-tarafa: I see some of this seems to have been [fixed](https://github.com/googleapis/gapic-showcase/commit/48cfc9b4417217dff3d6cb124313e82929d3fd21) on v0.44.0 and I see we default to v0.43.0 in some cases. Before I review, can you check that we are going against the newest Showcase version and that your regex to find latest is indeed working?

If the regex does not work or you get quota errors often, let's see if we can get the version as a constant on top of the script and renovate to update it for us?

robertvoinescu-work: Adding here again based on offline discussion. I've verified we are using the latest (now printing it out) and the issue is still not fixed. Will address the issue myself as a follow-up in the listed bug.

amanda-tarafa (approved):

